PR #19, opened 2017-09-06
add support for `connectionPollInterval ` and `maxConnectionRetries`

alekseykulikov:
This PR is a copy of https://github.com/GoogleChrome/lighthouse/pull/3129 and addresses #12 

Also this may fix https://github.com/GoogleChrome/lighthouse/pull/2616, because the main goal is to avoid `ECONNREFUSED`. In my experiments, this issue was related with not enough time in 100% cases, just increase from 5s to 25s fixed `ECONNREFUSED`.

### DIFF
--- a/chrome-launcher.ts
+++ b/chrome-launcher.ts
@@ -34,6 +34,8 @@ export interface Options {
   userDataDir?: string;
   logLevel?: string;
   enableExtensions?: boolean;
+  connectionPollInterval?: number;
+  maxConnectionRetries?: number;
 }
 
 export interface LaunchedChrome {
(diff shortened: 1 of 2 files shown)


chrome-launcher.ts
```diff
@@ -270,7 +275,7 @@ export class Launcher {
               resolve();
             })
             .catch(err => {
-              if (retries > 10) {
+              if (retries > launcher.maxConnectionRetries) {
```
samccone: Hi and thanks for the PR. I think this is a great addition!

alekseykulikov: Hi Sam, thanks!

PR #20, opened 2017-09-07
only include PROGRAMFILES(X86) if present

krystophv:
In order to address https://github.com/GoogleChrome/lighthouse/issues/3270

### DIFF
--- a/chrome-finder.ts
+++ b/chrome-finder.ts
@@ -155,11 +155,9 @@ export function win32() {
   const suffixes = [
     '\\Google\\Chrome SxS\\Application\\chrome.exe', '\\Google\\Chrome\\Application\\chrome.exe'
   ];
-  const prefixes = [process.env.LOCALAPPDATA, process.env.PROGRAMFILES];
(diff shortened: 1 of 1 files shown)


krystophv:
I signed it!

chrome-finder.ts
```diff
@@ -155,8 +155,11 @@ export function win32() {
   const suffixes = [
     '\\Google\\Chrome SxS\\Application\\chrome.exe', '\\Google\\Chrome\\Application\\chrome.exe'
   ];
-  const prefixes =
-      [process.env.LOCALAPPDATA, process.env.PROGRAMFILES, process.env['PROGRAMFILES(X86)']];
+  const prefixes = [process.env.LOCALAPPDATA, process.env.PROGRAMFILES];
```
paulirish: rather than this why not just do this at the end of the array:

```js
.filter(Boolean)
```

Helps in case we need to add one more env variable in the future?

krystophv: That's certainly a more general solution. Will change.

PR #23, opened 2017-09-12
Add chrome flags used in chromedriver/puppeteer

paulirish:
sources: 

https://github.com/GoogleChrome/puppeteer/commit/4af0911b90773e59c3ff552f6c4cc6e42af715a0
https://cs.chromium.org/chromium/src/chrome/test/chromedriver/chrome_launcher.cc?type=cs&q=f:chrome_launcher++kDesktopSwitches&sq=package:chromium&l=72

### DIFF
--- a/flags.ts
+++ b/flags.ts
@@ -6,15 +6,16 @@
 'use strict';
 
 export const DEFAULT_FLAGS = [
-  // Disable built-in Google Translate service
-  '--disable-translate',
   // Disable all chrome extensions entirely
   '--disable-extensions',
   // Disable various background network services, including extension updating,
   //   safe browsing service, upgrade detector, translate, UMA
   '--disable-background-networking',
-  // Disable fetching safebrowsing lists, likely redundant due to disable-background-networking
+  // Disable fetching safebrowsing lists. Otherwise requires a 500KB download immediately after launch.
+  // This flag is likely redundant due to disable-background-networking
   '--safebrowsing-disable-auto-update',
+  // Disable built-in Google Translate service
+  '--disable-translate',
   // Disable syncing to a Google account
   '--disable-sync',
   // Disable reporting to UMA, but allows for collection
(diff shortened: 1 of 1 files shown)


paulirish:
I also found `--test-type=foo` which has some interesting effects:  https://cs.chromium.org/search/?q=kTestType%5Cb&type=cs

Not sure about it yet though.

flags.ts
```diff
@@ -25,4 +26,20 @@ export const DEFAULT_FLAGS = [
   '--mute-audio',
   // Skip first run wizards
   '--no-first-run',
+
+  // Disable timers being throttled in background pages/tabs
+  '--disable-background-timer-throttling',
```
brendankenny: does this have any benefit for us? It seems like it could have an unrealistic effect on performance in certain corner cases of sites and it may be unnecessary for Lighthouse due to always being the foreground tab

paulirish: > unnecessary for Lighthouse due to always being the foreground tab

yes seems correct.

however since other projects use chrome-launcher... this would affect anyone attempting to use a pool of tabs to do a bunch of work.

flags.ts
```diff
@@ -25,4 +26,20 @@ export const DEFAULT_FLAGS = [
...
+  // Disables client-side phishing detection. Likely redundant due to the safebrowsing disable
+  '--disable-client-side-phishing-detection',
+  // Disable popup blocking
+  '--disable-popup-blocking',
+  // Reloading a page that came from a POST normally prompts the user.
+  '--disable-prompt-on-repost',
+  // Disable the a few things considered not appropriate for automation, e.g. infobar animations, password saving UI
```
patrickhulce: s/the a few/a few/

flags.ts
```diff
@@ -25,4 +26,20 @@ export const DEFAULT_FLAGS = [
...
+  //   https://docs.google.com/a/google.com/document/d/1JYj9K61UyxIYavR8_HATYIglR9T_rDwAtLLsD3fbDQg/preview
```
patrickhulce: this link doesn't actually reference that flag, do we have something that explains what it does? or is `--enable-automation` the same as `--enable-chromedriver`?

paulirish: Yah in short that design doc turned into --enable-automation during code review but the doc wasn't updated. 
CL here: https://codereview.chromium.org/2564973002 where pavel recommended the flag be generalized. :)

> My only comment is to call it in a generic manner: --enable-automation, AutomationInfobar, etc. It is likely that Lighthouse would also like to use it the same way ChromeDriver does.

it adds this infobar: 

image

and disables dev mode extension bubbles (?), and doesn't show some other info bars
and it means the default browser check prompt isn't shown
and it avoids showing these 3 infobars: ShowBadFlagsPrompt, GoogleApiKeysInfoBarDelegate, ObsoleteSystemInfoBarDelegate

JoelEinbinder:
I went over the differences between Puppeteer and the new chrome-launcher flags. The only differences are `--disable-translate` and `--disable-extensions`. I'm assuming translate might slow things down for Lighthouse. But extensions puzzles me. They shouldn't be loaded because a custom userDataDir is set right? And if the user manually specifies a userDataDir which has extensions, wouldn't they want them to run?

samccone:
happy to address the nit comment from @patrickhulce if everyone is happy to land this?

paulirish:
> happy to address the nit comment from @patrickhulce if everyone is happy to land this?

I think we need to have the flag disabling solution (#65) in place before landing this guy since there's a few ones in here some consumers of chrome-launcher may not want set by default.

PR #27, opened 2017-09-19
Add support for WSL platform

### LINKED ISSUE #26: Add support for WSL on Windows 10
Starting on [build 14951](https://msdn.microsoft.com/en-us/commandline/wsl/release_notes#build-14951) WSL supports opening windows applications from bash. This will be available for anyone in the next stable release of Windows in [October 17th](https://blogs.windows.com/windowsexperience/2017/09/01/create-and-play-this-holiday-with-the-windows-10-fall-creators-update-coming-oct-17/)

Bash on Windows is becoming very popular and adding support to this will be great. We've had a few people run into this issue when using [sonar](https://github.com/sonarwhal/sonar).

I can probably look into adding this if no one else wants to 😊

molant:
Fix #26

### DIFF
--- a/chrome-finder.ts
+++ b/chrome-finder.ts
@@ -11,6 +11,8 @@ const execSync = require('child_process').execSync;
 const execFileSync = require('child_process').execFileSync;
 const log = require('lighthouse-logger');
 
+import {getLocalAppDataPath} from './utils';
+
 const newLineRegex = /\r?\n/;
 
 type Priorities = Array<{regex: RegExp, weight: number}>;
@@ -150,10 +152,20 @@ export function linux() {
   return sort(uniq(installations.filter(Boolean)), priorities);
 }
 
+export function wsl() {
+  // Manually populate the environment variables assuming it's the default config
+  process.env.LOCALAPPDATA = getLocalAppDataPath();
+  process.env.PROGRAMFILES = '/mnt/c/Program Files';
+  process.env['PROGRAMFILES(X86)'] = '/mnt/c/Program Files (x86)';
+
+  return win32();
+}
+
 export function win32() {
   const installations: Array<string> = [];
   const suffixes = [
-    '\\Google\\Chrome SxS\\Application\\chrome.exe', '\\Google\\Chrome\\Application\\chrome.exe'
+    `${path.sep}Google${path.sep}Chrome SxS${path.sep}Application${path.sep}chrome.exe`,
+    `${path.sep}Google${path.sep}Chrome${path.sep}Application${path.sep}chrome.exe`
   ];
   const prefixes = [
     process.env.LOCALAPPDATA, process.env.PROGRAMFILES, process.env['PROGRAMFILES(X86)']
--- a/chrome-launcher.ts
+++ b/chrome-launcher.ts
@@ -12,16 +12,17 @@ import * as rimraf from 'rimraf';
 import * as chromeFinder from './chrome-finder';
 import {getRandomPort} from './random-port';
 import {DEFAULT_FLAGS} from './flags';
-import {makeTmpDir, defaults, delay} from './utils';
+import {makeTmpDir, defaults, delay, getPlatform, toWinDirFormat} from './utils';
 const log = require('lighthouse-logger');
 const spawn = childProcess.spawn;
 const execSync = childProcess.execSync;
-const isWindows = process.platform === 'win32';
+const isWsl = getPlatform() === 'wsl';
+const isWindows = getPlatform() === 'win32';
 const _SIGINT = 'SIGINT';
 const _SIGINT_EXIT_CODE = 130;
-const _SUPPORTED_PLATFORMS = new Set(['darwin', 'linux', 'win32']);
+const _SUPPORTED_PLATFORMS = new Set(['darwin', 'linux', 'win32', 'wsl']);
 
-type SupportedPlatforms = 'darwin'|'linux'|'win32';
+type SupportedPlatforms = 'darwin'|'linux'|'win32'|'wsl';
 
 const instances = new Set();
 
@@ -123,14 +124,15 @@ export class Launcher {
     let flags = DEFAULT_FLAGS.concat([
       `--remote-debugging-port=${this.port}`,
       // Place Chrome profile in a custom location we'll rm -rf later
-      `--user-data-dir=${this.userDataDir}`
+      // If in WSL, we need to use the Windows format
+      `--user-data-dir=${isWsl ? toWinDirFormat(this.userDataDir) : this.userDataDir}`
     ]);
 
     if (this.enableExtensions) {
       flags = flags.filter(flag => flag !== '--disable-extensions');
     }
 
-    if (process.platform === 'linux') {
+    if (getPlatform() === 'linux') {
       flags.push('--disable-setuid-sandbox');
     }
 
@@ -146,7 +148,7 @@ export class Launcher {
   }
 
   prepare() {
-    const platform = process.platform as SupportedPlatforms;
+    const platform = getPlatform() as SupportedPlatforms;
     if (!_SUPPORTED_PLATFORMS.has(platform)) {
       throw new Error(`Platform ${platform} is not supported`);
     }
@@ -178,17 +180,17 @@ export class Launcher {
       }
     }
 
-    if (!this.tmpDirandPidFileReady) {
-      this.prepare();
-    }
-
     if (this.chromePath === undefined) {
-      const installations = await chromeFinder[process.platform as SupportedPlatforms]();
+      const installations = await chromeFinder[getPlatform() as SupportedPlatforms]();
       if (installations.length === 0) {
         throw new Error('No Chrome Installations Found');
       }
 
       this.chromePath = installations[0];
+    }
+
+    if (!this.tmpDirandPidFileReady) {
+      this.prepare();
     }
 
     this.pid = await this.spawnProcess(this.chromePath);
--- a/utils.ts
+++ b/utils.ts
@@ -8,6 +8,7 @@
 import {join} from 'path';
 import {execSync} from 'child_process';
 import * as mkdirp from 'mkdirp';
+const isWsl = require('is-wsl');
 
 export function defaults<T>(val: T | undefined, def: T): T {
   return typeof val === 'undefined' ? def : val;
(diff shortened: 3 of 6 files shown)


utils.ts
```diff
@@ -9,26 +9,60 @@ import {join} from 'path';
...
 
+export function getPlatform() {
+  const platform = process.platform;
+
+  if (platform !== 'linux') {
+    return platform;
+  }
+  // Windows Subsystem for Linux, reports Linux as platform but chrome is on Windows
+  // we need to find out the real platform
+  if (process.env.PATH && (process.env.PATH as string).includes(':/mnt/c/Windows/System32:')) {
```
paulirish: this feels a little brittle.. (i forget if a windows install on non-C-drive is possible, but...)

I saw some other projects [that basically do](https://github.com/kairess/projectETTN/blob/8bdfbbe80f7330d92de3f8481c4fd8f99128262e/.meteor/local/build/programs/server/mini-files.js#L16-L22) `require('os').release().includes('Microsoft')`

but it looks like even that is incomplete: https://github.com/sindresorhus/is-wsl/commit/93b4fcb68f3a955335266c5757317ab680e9ff8f

maybe let's pull in is-wsl.

molant:
I signed it!

utils.ts
```diff
@@ -9,26 +9,60 @@ import {join} from 'path';
...
+
 export function makeTmpDir() {
-  switch (process.platform) {
+  switch (getPlatform()) {
     case 'darwin':
     case 'linux':
       return makeUnixTmpDir();
+    case 'wsl':
+      // We populate the user's Windows temp dir so the folder is correctly created later
+      process.env.TEMP = `/mnt/c/Users/${findWSLUser()}/AppData/Local/Temp`;
```
paulirish: This isn't set for WSL typically ?  Separately, I imagine `TMPDIR` is set, yeah? Can we use that'?

molant:
Not sure why it complains. Formatting should be ok. This is what I get when running `yarn format` and `yarn test-formatting`:

```
molant@DESKTOP-ANUL44G:~/Projects/chrome-launcher$ git push
Everything up-to-date
molant@DESKTOP-ANUL44G:~/Projects/chrome-launcher$ yarn format
yarn format v1.0.2
$ scripts/format.sh
Done in 0.80s.
molant@DESKTOP-ANUL44G:~/Projects/chrome-launcher$ yarn test-formatting
yarn test-formatting v1.0.2
$ test/check-formatting.sh
Done in 0.76s.
```

utils.ts
```diff
@@ -9,26 +9,60 @@ import {join} from 'path';
...
+export function toWinDirFormatIfNeeded(dir: string|any): string|any {
+  if (!dir || getPlatform() !== 'wsl') {
+    return dir;
+  }
+
+  return dir.replace('/mnt/c/', 'C:\\').replace(/\//g, '\\');
+}
+
+export function findWSLUser(): string {
+  const path = process.env.PATH;
```
paulirish: $USER isn't present? or maybe its possible it wont match the filepath?

utils.ts
```diff
@@ -9,26 +9,60 @@ import {join} from 'path';
...
 
+export function toWinDirFormatIfNeeded(dir: string|any): string|any {
+  if (!dir || getPlatform() !== 'wsl') {
+    return dir;
+  }
+
+  return dir.replace('/mnt/c/', 'C:\\').replace(/\//g, '\\');
+}
+
+export function findWSLUser(): string {
```
paulirish: Everywhere this is used, it's mostly just to resolve the LOCALAPPDATA path. how about we change this method to be a `getLocalAppDataPath()`?

chrome-finder.ts
```diff
@@ -150,10 +152,23 @@ export function linux() {
...
+
+  return win32();
+}
+
 export function win32() {
   const installations: Array<string> = [];
+  // Because we could be in WSL, we need to make sure to use the right separator
   const suffixes = [
-    '\\Google\\Chrome SxS\\Application\\chrome.exe', '\\Google\\Chrome\\Application\\chrome.exe'
+    `${path.sep}Google${path.sep}Chrome SxS${path.sep}Application${path.sep}chrome.exe`,
```
paulirish: kinda awkward. how about using `path.join()` instead?

chrome-finder.ts
```diff
@@ -150,10 +152,23 @@ export function linux() {
...
+  process.env.LOCALAPPDATA = `/mnt/c/Users/${user}/AppData/Local`;
+  process.env.PROGRAMFILES = '/mnt/c/Program Files';
+  process.env['PROGRAMFILES(X86)'] = '/mnt/c/Program Files (x86)';
+
+  return win32();
+}
+
 export function win32() {
   const installations: Array<string> = [];
+  // Because we could be in WSL, we need to make sure to use the right separator
```
paulirish: you can drop this comment IMO

utils.ts
```diff
@@ -9,26 +9,60 @@ import {join} from 'path';
...
     case 'darwin':
     case 'linux':
       return makeUnixTmpDir();
+    case 'wsl':
+      // We populate the user's Windows temp dir so the folder is correctly created later
+      process.env.TEMP = `/mnt/c/Users/${findWSLUser()}/AppData/Local/Temp`;
     case 'win32':
       return makeWin32TmpDir();
     default:
       throw new Error(`Platform ${process.platform} is not supported`);
```
paulirish: use getPlatform() instead

chrome-launcher.ts
```diff
@@ -12,16 +12,16 @@ import * as rimraf from 'rimraf';
 import * as chromeFinder from './chrome-finder';
 import {getRandomPort} from './random-port';
 import {DEFAULT_FLAGS} from './flags';
-import {makeTmpDir, defaults, delay} from './utils';
+import {makeTmpDir, defaults, delay, getPlatform, toWinDirFormatIfNeeded} from './utils';
 const log = require('lighthouse-logger');
 const spawn = childProcess.spawn;
 const execSync = childProcess.execSync;
 const isWindows = process.platform === 'win32';
```
paulirish: use getPlatform() instead. also on L134 and L150

chrome-launcher.ts
```diff
@@ -123,7 +123,8 @@ export class Launcher {
     let flags = DEFAULT_FLAGS.concat([
       `--remote-debugging-port=${this.port}`,
       // Place Chrome profile in a custom location we'll rm -rf later
-      `--user-data-dir=${this.userDataDir}`
+      // If in WSL, we need to use the Windows format
+      `--user-data-dir=${toWinDirFormatIfNeeded(this.userDataDir)}`
```
paulirish: looks like the other uses of userDataDir don't need the path to be rewritten?

utils.ts
```diff
@@ -9,26 +9,60 @@ import {join} from 'path';
...
   }
 }
 
+export function toWinDirFormatIfNeeded(dir: string|any): string|any {
+  if (!dir || getPlatform() !== 'wsl') {
+    return dir;
+  }
+
+  return dir.replace('/mnt/c/', 'C:\\').replace(/\//g, '\\');
```
paulirish: we probably could use a test for this method.

paulirish:
Formatting. this line is the difference: 

```diff
-export function defaults<T>(val: T|undefined, def: T): T {
+export function defaults<T>(val: T | undefined, def: T): T {
```

chrome-finder.ts
brendankenny: Sorry if I'm missing some context, but windows will happily use `\` or `/` as directory separators (it's only path separators (`:` vs `;`) that are different). So this might not be necessary?

chrome-finder.ts
molant: Done

chrome-finder.ts
molant: That was my initial though but it didn't work. If I don't change things I end up with the following paths:
```
/mnt/c/Users/antonmo/AppData/Local/\Google\Chrome SxS\Application\chrome.exe
/mnt/c/Users/antonmo/AppData/Local/\Google\Chrome\Application\chrome.exe
/mnt/c/Program Files/\Google\Chrome SxS\Application\chrome.exe
/mnt/c/Program Files/\Google\Chrome\Application\chrome.exe
/mnt/c/Program Files (x86)/\Google\Chrome SxS\Application\chrome.exe
/mnt/c/Program Files (x86)/\Google\Chrome\Application\chrome.exe
```

and no installations are found:
```
(node:3007) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: No Chrome Installations Found
(node:3007) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
I tried doing `path.normalize` when joining with the prefix (`const chromePath = path.normalize(path.join(prefix, suffix));`) but the result was still the same (at least on 8.4.0) :(

chrome-launcher.ts
molant: Done

chrome-launcher.ts
molant: No, this is only for WSL. The reason is that WSL access that path via `/mnt/c/...` but then we execute chrome in windows land and it doesn't recognize that path format at all so we need to convert it to `C:\...`

utils.ts
molant: Awesome, didn't know about it. I'm doing a `require` because of:
> file: 'file:///c%3A/Projects/chrome-launcher/utils.ts'
severity: 'Error'
message: 'Could not find a declaration file for module 'is-wsl'. 'c:/Projects/chrome-launcher/node_modules/is-wsl/index.js' implicitly has an 'any' type.
  Try `npm install @types/is-wsl` if it exists or add a new declaration (.d.ts) file containing `declare module 'is-wsl';`'
at: '11,24'
source: 'ts'

I could add the `.d.ts.` file but wasn't sure what you prefer.

utils.ts
molant: Unfortunately it isn't set. Also because we are on Linux the tempdir is different than the Windows one. and Chrome (from Windows) can't access the WSL tempdir.
RE the user name, it can be different between Windows and Linux. I spent a few minutes trying to figure out what was wrong on my machine because of that :(

utils.ts
molant: $USER can be different between WSL and Windows

utils.ts
molant: Done

molant:
@paulirish PR updated, tests passing and added a few comments in GitHub explaining some things.
I'll squash once everything looks good to you.

Thanks!

paulirish:
shipped in 0.8.0

molant:
Thanks 🎉

PR #33, opened 2017-09-28
Move types to devDependencies

remcohaszing:
These type definitions exist for the typescript compiler, not for usage
at runtime. Users of this package don't need these.

This makes them devDependencies, not dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,11 @@
     "format": "scripts/format.sh"
   },
   "devDependencies": {
+    "@types/core-js": "^0.9.41",
+    "@types/mkdirp": "^0.3.29",
     "@types/mocha": "^2.2.41",
+    "@types/node": "6.0.66",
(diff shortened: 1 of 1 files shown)


patrickhulce:
thanks for the contribution @remcohaszing! 

this was actually intentional though so that typescript consumers of this package do not need to install the type definitions themselves in order to successfully compile (see https://github.com/Microsoft/types-publisher/issues/81#issuecomment-234051338)

AFAIK we don't expose these signatures, so if consumers don't need to re-compile our code for a lower target then they are in fact unnecessary, but the ergonomic convenience is nice to support those cases seamlessly.

samccone:
If i recall correctly... when someone was consuming the TS directly the compile would fail until they installed the typings. That is why we moved them to the deps.

paulirish:
based on the two comments in types-publisher i agree these should remain in deps.

thx for starting the discussing tho

PR #36, opened 2017-10-05
Add more paths for Chromium

kissifrot:
This adds more paths for Chromium

`\\Chromium\\Application\\chrome.exe` used by installer at https://chromium.woolyss.com/
`\\chrome-win32\\chrome.exe` used if unzipping file downloaded from https://download-chromium.appspot.com/?platform=Win&type=snapshots

### DIFF
--- a/chrome-finder.ts
+++ b/chrome-finder.ts
@@ -33,7 +33,7 @@ export function darwin() {
 
   execSync(
       `${LSREGISTER} -dump` +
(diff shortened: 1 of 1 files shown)


kissifrot:
I signed it!

paulirish:
Since `CHROME_PATH` is configurable i'd rather support these uncommon cases that way.

the automagic of chrome_finder is a double edged sword. :)

PR #38, opened 2017-10-22
add enable audio option fix #28

### LINKED ISSUE #28: Allow mute-audio flag to be disabled
I understand the majority of users will want muted audio during automated testing, but in our case we are recording video of the browsers tests and would like to record any audio generated by the browser.  I'm sure others with multimedia sites may need access to this functionality.  Although audio can't easily be automatically tested for, it is useful and simpler to watch the automated test case videos to check functionality rather than having to manually test audio functionality separately.
 
As this flag is in the defaults, there appears to be no way to override the behaviour (short of forking the entire code). 

Would it be possible to make this a non-default flag, as an option that defaults to true, or an --unmute-audio flag, that pops the --mute-audio from the defaults?

Awarua-:
This pull request adds an enableAudio flag to the options of the chrome launcher.

Fixes #28 

Although I agree that audio should be disabled by default.
For my use case of the chrome-launcher, I require audio to be active.

I am unsure whether the option should be named enableAudio or unmute.

I feel like enableAudio is clearer in what it does, but unmute is more in line with the --mute-audio flag.

I suppose you could compromise and name the option unmuteAudio? 😕

### DIFF
--- a/chrome-launcher.ts
+++ b/chrome-launcher.ts
@@ -34,7 +34,7 @@ export interface Options {
   chromePath?: string;
   userDataDir?: string;
   logLevel?: string;
-  enableExtensions?: boolean;
+  ignoreDefaultFlags?: boolean;
   connectionPollInterval?: number;
   maxConnectionRetries?: number;
 }
@@ -89,7 +89,7 @@ export class Launcher {
   private outFile?: number;
   private errFile?: number;
   private chromePath?: string;
-  private enableExtensions?: boolean;
+  private ignoreDefaultFlags?: boolean;
   private chromeFlags: string[];
   private requestedPort?: number;
   private connectionPollInterval: number;
@@ -115,22 +115,19 @@ export class Launcher {
     this.chromeFlags = defaults(this.opts.chromeFlags, []);
     this.requestedPort = defaults(this.opts.port, 0);
     this.chromePath = this.opts.chromePath;
-    this.enableExtensions = defaults(this.opts.enableExtensions, false);
+    this.ignoreDefaultFlags = defaults(this.opts.ignoreDefaultFlags, false);
     this.connectionPollInterval = defaults(this.opts.connectionPollInterval, 500);
     this.maxConnectionRetries = defaults(this.opts.maxConnectionRetries, 50);
   }
 
   private get flags() {
-    let flags = DEFAULT_FLAGS.concat([
+    let flags = this.ignoreDefaultFlags ? [] : DEFAULT_FLAGS;
+    flags = flags.concat([
       `--remote-debugging-port=${this.port}`,
       // Place Chrome profile in a custom location we'll rm -rf later
       // If in WSL, we need to use the Windows format
       `--user-data-dir=${isWsl ? toWinDirFormat(this.userDataDir) : this.userDataDir}`
     ]);
-
-    if (this.enableExtensions) {
-      flags = flags.filter(flag => flag !== '--disable-extensions');
-    }
 
     if (getPlatform() === 'linux') {
       flags.push('--disable-setuid-sandbox');
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ npm install chrome-launcher
   // Default: 'info'
   logLevel: string;
 
-  // (optional) Enable extension loading
+  // (optional) Flags specific in [flags.ts](https://github.com/GoogleChrome/chrome-launcher/blob/master/flags.ts) will not be included.
   // Default: false
-  enableExtensions: boolean;
+  ignoreDefaultFlags: booelan;
 
   // (optional) Interval in ms, which defines how often launcher checks browser port to be ready.
   // Default: 500
--- a/test/chrome-launcher-test.ts
+++ b/test/chrome-launcher-test.ts
@@ -16,6 +16,23 @@ const fsMock = {
   writeFileSync: () => {}
 };
 
+const launchChromeWithOpts = async (opts: object) => {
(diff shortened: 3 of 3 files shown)


Awarua-:
I signed it!

paulirish:
I'm thinking about this sort of API rather than creating individual disable-the-disable flags

https://github.com/GoogleChrome/puppeteer/issues/872

how does that look to you?

Awarua-:
Yes, I agree that is a better solution, 😄

I have changed my commit to reflect this.
Although I am unsure if defaultIgnoreFlags is the best name 😕

chrome-launcher.ts
```diff
@@ -128,9 +128,12 @@ export class Launcher {
       `--user-data-dir=${isWsl ? toWinDirFormat(this.userDataDir) : this.userDataDir}`
     ]);
 
-    if (this.enableExtensions) {
-      flags = flags.filter(flag => flag !== '--disable-extensions');
-    }
+    flags = flags.filter(flag => {
+      if (this.defaultIgnoreFlags) {
```
Awarua-: TypeScript complains if this if statement is missing.
Due to strict null type checking

paulirish:
@ragingwind wdyt about this new API?   

we could kill enableExtensions now or deprecate now and kill for real in the future.

paulirish:
@Awarua- there's a bit of a difference between what you have here and what https://github.com/GoogleChrome/puppeteer/issues/872 proposes..

i was saying something like this:

```js
chromeLauncher.defaultFlags() // array of all the flags it uses by default

chromeLauncher.launch({
   ignoreDefaultFlags: true,  // will not supply any defaults.. the ones passed in below are it.
  chromeFlags: [] // array of the only flags to use
});
```

that said, nice commits! :) appreciate the updates in the tests.

Awarua-:
@paulirish Oh yeah, I see what you mean.

Which method would you prefer, the current implementation or the one in https://github.com/GoogleChrome/puppeteer/issues/872

ragingwind:
@paulirish LGTM. I was also considered how to manage disable-flags

paulirish:
@Awarua- lets do what i spelled out in this comment: https://github.com/GoogleChrome/chrome-launcher/pull/38#issuecomment-338516008

Awarua-:
@paulirish Changes made as requested 🤞

paulirish:
This is really nice. i'll probably merge but at a conference right now so give me 2 days

brendankenny:
This (and the approach in https://github.com/GoogleChrome/puppeteer/issues/872) works well for access as a node module, but I think this was the same CLI problem we faced when we ended up adding `--enable-extensions`. If that's the only flag you want to change, there's no easy way to keep all the default flags except `--disable-extensions`

brendankenny:
Oh, I guess my comment only applies to [`manual-chrome-launcher`](https://github.com/GoogleChrome/chrome-launcher/blob/master/manual-chrome-launcher.js) (here and in lighthouse). Why don't we expose a `bin` entry for `chrome-launcher`?

Zarlex:
But now all default flags are gone when  this option is set to true. What is when i only want to remove only one default flag? In this case i would have to add all other flags manually. 
So in your scenario you want to remove only the `mute-audio` flag. But you probably want to have the other default flags, so you have to add them manually to the `chromeFlags`. You also have to maintain them manually in case other default flags are added in a future version.
I was thinking about maybe having an option attribute like  `ignoreDefaultsFlags=["--mute-audio"]` and those attributes are removed from the default flags. What do you think about that?

Awarua-:
@Zarlex 
A previous implementation of this PR had the exact implementation that you have described, where there is an optional list of flags that are removed from the default flags.
However, @paulirish suggested we adhere to the API described in this issue https://github.com/GoogleChrome/puppeteer/issues/872

I guess it's up to the maintainers of the project to decide what they want.

For me, I'm not using chrome launcher as part of a CI pipeline, but instead, I am using it for HCI experiments which involves launching multiple instances of chrome of separate displays. So my use case is very niche.

Either way, I'm not worried if I have to add the default flags manually to chromeflags as my experiment code has a short lifespan. But others may have different concerns/use cases 😕

ragingwind:
It's not easy but I think we can manage flags with pattern match

```
// enable all of flags
chrome --flags

// enable a flags
chrome --flags=extension,audio

**Comment by Awarua-** [comment]
@ragingwind I think I understand what you are meaning.
But I don't understand what this would look like in terms of launchArgs
Can you provide an example?

**Comment by ragingwind** [comment]
@Awarua- My comment about supporting command options for CLI launcher. It will be transformed to ignoreFlags array

**Comment by paulirish** [comment]
in the lighthouse CLI was have `--list-all-audits` and `--list-trace-categories`, so I think we'd do something similar here (basically what @ragingwind was saying)

so for a launcher CLI (either with ragingwind's module or our manual-chrome-launcher.js):

```sh
chrome-launcher --list-flags # prints and quits: '--disable-extensions --disable-translate-service ....'

chrome-launcher # launches with default flags
chrome-launcher --set-flags="--show-paint-rects --disable-extensions"   # skips default flags and only uses provided ones
```

@brendankenny  wdyt

samccone:
I am happy with this implementation :) @paulirish if you are also I am happy to land. 

This is going to require a major semver bump of the package as it is breaking but I think it is the correct move forward.

Thanks so much for the PR @Awarua- and sorry about the delay!

samccone:
Hey @Awarua- thanks again for your PR I went ahead and integrated your changes here
https://github.com/GoogleChrome/chrome-launcher/pull/70

Thanks!!!

PR #40, opened 2017-10-22
Some users may want access to the underlying child process

ORESoftware:
Some users may want/need access to the underlying child process


ORESoftware:
Please either pass `instance`, or pass `instance.chrome`

That would help me

ORESoftware:
@google-admin could use help with this, definitely looking to get this resolved one way or another, thx

patrickhulce:
I believe there was desire to minimize the API surface users came to rely on so that our implementation could stay flexible. Exposing the `instance.chrome` property for the child process seems reasonable though.

@paulirish any strong opinions here?

the1mills:
Exposing instance is not necessary, but exposing instance.chrome is imperative to my use case.

But I think exposing instance itself should be fine if you want to cast a wider net for more users use cases.

shi-cong:
i have tested , i run a http server and i use chrome headless send http request to my server, i print the request headers , and find chrome headless user agent in "user-agent" param, so , chrome headless is just for testers not for all.

ORESoftware:
@shi-cong not sure if your comment was intended for this thread?

ORESoftware:
@google-admin please either merge this or explain what needs fixing

patrickhulce:
@ORESoftware it's been established that we are resistant to exposing the entire Launcher instance to avoid reliance on an API we expect to be able to change.

Is there a reason exposing a `process: instance.chrome` property is not sufficient?

ORESoftware:
`instance.chrome` property is indeed sufficient, do you want me to make a new PR?

chrome-launcher.ts
```diff
@@ -79,7 +79,7 @@ export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
     return instance.kill();
   };
 
-  return {pid: instance.pid!, port: instance.port!, kill};
+  return {pid: instance.pid!, port: instance.port!, kill, proc: instance.chrome};
```
patrickhulce: is there a particular reason we're exposing under `.proc` instead of `.process` or `.processInstance` or something? 

I'm totally fine if there's precedent and we're matching what several other node libraries expose, but if not, I'd prefer clarity

ORESoftware: I was worried I'd  get some nitpicks on that, I don't care what we call it. We can call it `chromeProcess` if you want?

ORESoftware: Do you know why the build failed?

patrickhulce: `chromeProcess` sgtm 👍 

and yes, `.chrome` is marked as private in typescript you'll need to move the declaration in `class Launcher` to be with the other public properties

ORESoftware: ok cool, np, I just updated the PR, hopefully it will take thx

ORESoftware:
@patrickhulce this should be good now, please advise, thx

ORESoftware:
pls merge this thx :)

chrome-launcher.ts
```diff
@@ -79,7 +79,7 @@ export async function launch(opts: Options = {}): Promise<LaunchedChrome> {
     return instance.kill();
   };
 
-  return {pid: instance.pid!, port: instance.port!, kill};
+  return {pid: instance.pid!, port: instance.port!, kill, chromeProcess: instance.chrome};
```
paulirish: can you add a test to assert this chromeProcess prop exists on the return value.. and that it's an instanceof childProcess?

ORESoftware: you want an assertion in the codebase or a unit test?

paulirish: hah good question. a unit test. :)  glanced quick and we don't have anything which asserts the shape of the return object API, but that generally makes sense to me.

ORESoftware: I added a unit test for the launch helper function return signature.

also, there was no .gitignore file, so node_modules is committed, fyi.

ORESoftware: I am going to assume that you want me to ignore node_modules...lmk

ORESoftware: Looks like there are a lot of changes to .js and .d.ts files, should we add those to .gitignore

samccone:
@ORESoftware thanks so much for the PR. I went ahead and finished this one up for you!

Thanks a bunch!!!!
https://github.com/GoogleChrome/chrome-launcher/pull/67

ORESoftware:
awesome blossom thanks that will help a lot

paulirish:
This functionality was landed and shipped in 0.10.0

PR #48, opened 2017-11-07
New: Add flag to use default user profile instead

### LINKED ISSUE #47: How to use default userDataDir on launch?
Is it possible to launch chrome without creating a new `userDataDir`?
From what I read in the documentation I will have to pass the user's current `userDataDir` as part of the launch options but that doesn't look OK to me. The developer will have to detect the platform and user name to generate the right path (`C:\Users\<username>\AppData\Local\Google\Chrome\User Data\Default`, `Users/<username>/Library/Application Support/Google/Chrome/Default`, `/home/<username>/.config/google-chrome/default`, for example).

The scenario I have in mind is to easily allow automation of some pages behind authentication if the user is already authenticated in there, for example.

Will you be open to have something to enable this? If so, what would be your preferred option? New parameter?  Special value in `userDataDir` (maybe `default`)?

molant:
Add a new property `defaultProfile` to use the default user's profile in Chrome instead of creating a new one. By default this value is `false` so it doesn't change the current behavior.

Close #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ npm install chrome-launcher
   // * Otherwise, a detected Chrome (stable) will be used
   chromePath: string;
 
-  // (optional) Chrome profile path to use
+  // (optional) Chrome profile path to use, if set to `false` then the default profile will be used.
   // By default, a fresh Chrome profile will be created
-  userDataDir: string;
+  userDataDir: string | boolean;
 
   // (optional) Starting URL to open the browser with
   // Default: `about:blank`
--- a/chrome-launcher.ts
+++ b/chrome-launcher.ts
@@ -32,7 +32,7 @@ export interface Options {
   port?: number;
   handleSIGINT?: boolean;
   chromePath?: string;
-  userDataDir?: string;
+  userDataDir?: string|boolean;
   logLevel?: string;
   enableExtensions?: boolean;
   connectionPollInterval?: number;
@@ -98,6 +98,7 @@ export class Launcher {
   private fs: typeof fs;
   private rimraf: typeof rimraf;
   private spawn: typeof childProcess.spawn;
+  private useDefaultProfile: boolean;
 
   userDataDir?: string;
   port?: number;
@@ -118,15 +119,28 @@ export class Launcher {
     this.enableExtensions = defaults(this.opts.enableExtensions, false);
     this.connectionPollInterval = defaults(this.opts.connectionPollInterval, 500);
     this.maxConnectionRetries = defaults(this.opts.maxConnectionRetries, 50);
+
+    if (typeof this.opts.userDataDir === 'boolean') {
+      if (!this.opts.userDataDir) {
+        this.useDefaultProfile = true;
+        this.userDataDir = undefined;
+      } else {
+        throw new Error('userDataDir must be false or a path');
+      }
+    } else {
+      this.useDefaultProfile = false;
+      this.userDataDir = this.opts.userDataDir;
+    }
   }
 
   private get flags() {
-    let flags = DEFAULT_FLAGS.concat([
-      `--remote-debugging-port=${this.port}`,
-      // Place Chrome profile in a custom location we'll rm -rf later
+    let flags = DEFAULT_FLAGS.concat([`--remote-debugging-port=${this.port}`]);
+
+    // Place Chrome profile in a custom location we'll rm -rf later
+    if (!this.useDefaultProfile) {
       // If in WSL, we need to use the Windows format
-      `--user-data-dir=${isWsl ? toWinDirFormat(this.userDataDir) : this.userDataDir}`
-    ]);
+      flags.push(`--user-data-dir=${isWsl ? toWinDirFormat(this.userDataDir) : this.userDataDir}`);
+    }
 
     if (this.enableExtensions) {
       flags = flags.filter(flag => flag !== '--disable-extensions');
@@ -153,7 +167,7 @@ export class Launcher {
       throw new Error(`Platform ${platform} is not supported`);
     }
 
-    this.userDataDir = this.opts.userDataDir || this.makeTmpDir();
+    this.userDataDir = this.userDataDir || this.makeTmpDir();
     this.outFile = this.fs.openSync(`${this.userDataDir}/chrome-out.log`, 'a');
     this.errFile = this.fs.openSync(`${this.userDataDir}/chrome-err.log`, 'a');
 
--- a/test/chrome-launcher-test.ts
+++ b/test/chrome-launcher-test.ts
@@ -123,6 +123,25 @@ describe('Launcher', () => {
     assert.ok(!chromeFlags.includes('--disable-extensions'));
   });
 
+  it('removes --user-data-dir if userDataDir is false', async () => {
+    const spawnStub = stub().returns({pid: 'some_pid'});
+
+    const chromeInstance = new Launcher(
+        {userDataDir: false}, {fs: fsMock as any, rimraf: spy() as any, spawn: spawnStub as any});
+    stub(chromeInstance, 'waitUntilReady').returns(Promise.resolve());
+
+    chromeInstance.prepare();
+
+    try {
+      await chromeInstance.launch();
+    } catch (err) {
+      return Promise.reject(err);
+    }
+
+    const chromeFlags = spawnStub.getCall(0).args[1] as string[];
+    assert.ok(!chromeFlags.includes('--user-data-dir'));
+  });
+
   it('throws an error when chromePath is empty', (done) => {
     const chromeInstance = new Launcher({chromePath: ''});
     chromeInstance.launch().catch(() => done());


chrome-launcher.ts
```diff
@@ -118,15 +120,17 @@ export class Launcher {
...
     this.maxConnectionRetries = defaults(this.opts.maxConnectionRetries, 50);
+    this.defaultProfile = defaults(this.opts.defaultProfile, false);
   }
 
   private get flags() {
-    let flags = DEFAULT_FLAGS.concat([
-      `--remote-debugging-port=${this.port}`,
+    let flags = DEFAULT_FLAGS.concat([`--remote-debugging-port=${this.port}`]);
+
+    if (!this.defaultProfile) {
```
patrickhulce: let's also update the logic for conditionally creating/destroying the tmp directory

README.md
```diff
@@ -50,6 +50,10 @@ npm install chrome-launcher
   // * Otherwise, a detected Chrome (stable) will be used
   chromePath: string;
 
+  // (optional) Uses the default Chrome profile instead of creating a new one
+  // Default: false
+  defaultProfile: boolean;
```
patrickhulce: to 🚲 🏠  on name, it's not super clear that this will currently ignore whatever is set as `userDataDir`

some potential others

* `forceDefaultProfile`
* `ignoreUserDataDir`
* `overrideUserDataDirWithDefault`

alternatively, we could make this the behavior when `false` is explicitly passed to `userDataDir` and just document that it results in the default profile being used?

molant: I like `forceDefaultProfile`. 

> alternatively, we could make this the behavior when `false` is explicitly passed to `userDataDir` and just document that it results in the default profile being used?

Do you want me to add this behavior on top of the current one or instead of the new parameter? `userDataDir` will probably be a cleaner API for the user, but I was looking into the code and I had to start changing where the `pid` files were created, etc. It was starting to get a bit too complicated so I preferred to wait for feedback.

chrome-launcher.ts
molant: I'm still using a tmp directory for the logs and the pid file. If I don't use one, should I write those files into the default profile? It can get messy because then I'll have to guess the path of the default profile and delete each file individually.

README.md
patrickhulce: > Do you want me to add this behavior on top of the current one or instead of the new parameter? 

My personal preference would probably be the boolean approach instead of the new flag. I agree it's a cleaner API for consumers. OTOH, `forceDefaultProfile` is fairly explicit while `userDataDir: false` is a little more opaque even if the logic tracks of "do not pass in --user-data-dir".

> userDataDir will probably be a cleaner API for the user, but I was looking into the code and I had to start changing where the pid files were created, etc. It was starting to get a bit too complicated so I preferred to wait for feedback.

Hmmm, I suppose there's not a clear and obvious alternative for where else to stick the pid file `process.cwd()` might be ok but permissions and litter could get messy. Keeping the tmp directory for now seems ok to me.

molant: > OTOH, `forceDefaultProfile` is fairly explicit while `userDataDir: false` is a little more opaque even if the logic tracks of "do not pass in --user-data-dir".

Been trying to do this and there are some things to take into account. The main problem is "what happens if the users puts `true` instead of a `path` or `false`? Also I need to update the logic in a couple places to validate that `userDataDir` is a `string` instead of `boolean`, otherwise TypeScript will complain.

This is my current code in the constructor:

```js
if (typeof this.opts.userDataDir === 'boolean') {
      if (!this.opts.userDataDir) {
        this.forceDefaultProfile = true;
      } else {
        throw new Error('userDataDir has to be false or a path');
      }
    } else {
      this.forceDefaultProfile = defaults(this.opts.forceDefaultProfile, false);
    }
  }
```

What do you think? Should I just add the `forceDefaultProfile` parameter or continue down this path?

molant:
@patrickhulce I've created a new PR that just changes the property to `forceDefaultProfile` in https://github.com/GoogleChrome/chrome-launcher/pull/49
And updated this one to allow `userDataString` to be `false`. Let me know what do you prefer.

patrickhulce:
I really like the ergonomics of `userDataDir: false`, but feels like we need a bit of cleanup internally to reflect that the tmp dir is now the `logDir`, etc

lemme take the temperature of some other folks before having you go down the rabbit hole if I'm alone in preferring this over #49 :)

@brendankenny @samccone any preferences on this?

chrome-launcher.ts
```diff
@@ -337,7 +350,9 @@ export class Launcher {
         delete this.errFile;
       }
 
-      this.rimraf(this.userDataDir, () => resolve());
+      if (typeof this.userDataDir === 'string') {
```
brendankenny: maybe just check `this.forceDefaultProfile`?

chrome-launcher.ts
```diff
@@ -118,15 +120,26 @@ export class Launcher {
     this.enableExtensions = defaults(this.opts.enableExtensions, false);
     this.connectionPollInterval = defaults(this.opts.connectionPollInterval, 500);
     this.maxConnectionRetries = defaults(this.opts.maxConnectionRetries, 50);
+
+    if (typeof this.opts.userDataDir === 'boolean') {
+      if (!this.opts.userDataDir) {
+        this.forceDefaultProfile = true;
```
brendankenny: bikeshed: if it's a private property only, something like `useDefaultProfile` is another possible name

chrome-launcher.ts
```diff
@@ -118,15 +120,26 @@ export class Launcher {
...
+      } else {
+        throw new Error('userDataDir has to be false or a path');
```
brendankenny: nit: maybe `userDataDir must be false or a path`

chrome-launcher.ts
molant: Need to do that check or otherwise TypeScript will complain it can be `Boolean` or `string`

chrome-launcher.ts
molant: OK

molant:
> Can we drop forceDefaultProfile as an external option at the same time?

Sure, I'll close the other PR and update this one.

molant:
OK, I've updated the PR with @brendankenny's feedback. Once everyone is OK I'll squash the commits.

Let me know if there is something else you want me to do here!

README.md
```diff
@@ -50,9 +50,10 @@ npm install chrome-launcher
   // * Otherwise, a detected Chrome (stable) will be used
   chromePath: string;
 
-  // (optional) Chrome profile path to use
+    // (optional) Chrome profile path to use, if set to `false` then the default profile will be used.
+  // This is the same as setting `forceDefaultProfile` to `true`
```
patrickhulce: let's remove this now that `forceDefaultProfile` isn't an option anymore

chrome-launcher.ts
```diff
@@ -33,6 +33,7 @@ export interface Options {
   handleSIGINT?: boolean;
   chromePath?: string;
   userDataDir?: string;
+  forceDefaultProfile?: boolean;
```
patrickhulce: remove this now since we're using `useDefaultProfile` for the same purpose?

chrome-launcher.ts
patrickhulce: well we can always keep userDataDir a string inside this class, it's just `boolean|string` in `this.opts`

chrome-launcher.ts
```diff
@@ -118,15 +120,26 @@ export class Launcher {
...
     this.maxConnectionRetries = defaults(this.opts.maxConnectionRetries, 50);
+
+    if (typeof this.opts.userDataDir === 'boolean') {
+      if (!this.opts.userDataDir) {
+        this.useDefaultProfile = true;
+      } else {
+        throw new Error('userDataDir must be false or a path');
+      }
+    } else {
+      this.useDefaultProfile = defaults(this.opts.forceDefaultProfile, false);
```
patrickhulce: do we need this now? I guess we could explicitly set `this.useDefaultProfile` to false now that forceDefaultProfile is gone

molant:
I'll take a look at the tests on Monday. Thanks for the feedback!

molant:
@patrickhulce I've updated the tests. Had to add a new `dev-dependency` to mock the `makeTmpDir` because I've moved that part to the constructor so I didn't have to copy paste the condition again in `prepare`.

Let me know if I should change something else.

Thanks!

molant:
> so that merely constructing ChromeLauncher wouldn't have any side effects

[I'm just setting the value of `userDataDir`](https://github.com/GoogleChrome/chrome-launcher/pull/48/files#diff-d0585a92262648e2c52a750529099d64R132), no tmp directory gets created.

Do you still want me to change it?

brendankenny:
> I'm just setting the value of userDataDir, no tmp directory gets created.

doesn't the `this.makeTmpDir()` call do that?

molant:
You are right, I thought the magic was happening in [this.fs.openSync](https://github.com/GoogleChrome/chrome-launcher/blob/b08cd10b1f458f6280edcde6aec4249446468c7f/chrome-launcher.ts#L157)

Will change it and try to make it DRY

molant:
@brendankenny @patrickhulce 🤞

chrome-launcher.ts
```diff
@@ -121,12 +122,13 @@ export class Launcher {
   }
 
   private get flags() {
-    let flags = DEFAULT_FLAGS.concat([
-      `--remote-debugging-port=${this.port}`,
+    let flags = DEFAULT_FLAGS.concat([`--remote-debugging-port=${this.port}`]);
+
+    if (!this.useDefaultProfile) {
       // Place Chrome profile in a custom location we'll rm -rf later
```
patrickhulce: let's maybe move this line of the comment down to where we create it with a bit of explanation on the new logic

yarn.lock
```diff
@@ -1,1494 +1,1494 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
patrickhulce: this just seems like noise without any new deps, would you mind reverting this?

chrome-launcher.ts
```diff
@@ -153,7 +155,18 @@ export class Launcher {
       throw new Error(`Platform ${platform} is not supported`);
     }
 
-    this.userDataDir = this.opts.userDataDir || this.makeTmpDir();
+    if (typeof this.opts.userDataDir === 'boolean') {
```
brendankenny: maybe set `this.useDefaultProfile` in the profile?

Then in the constructor it would just be something like
```js
if (typeof this.opts.userDataDir === 'boolean') {
  if (!this.opts.userDataDir) {
    this.useDefaultProfile = true;
  } else {
    throw new Error('userDataDir must be false or a path');
  }
} else {
  this.useDefaultProfile = false;
}
```

That would immediately thrown on an invalid `true` for `userDataDir`, and it means that this line could just be 
`this.userDataDir = this.opts.userDataDir || this.makeTmpDir();`

brendankenny: that might be too clever for the typescript compiler, I guess. To go with @patrickhulce's suggestion to keep `this.userDataDir` an optional string, in the constructor could do

```js
if (typeof this.opts.userDataDir === 'boolean') {
  if (!this.opts.userDataDir) {
    this.useDefaultProfile = true;
    this.userDataDir = undefined;
  } else {
    throw new Error('userDataDir must be false or a path');
  }
} else {
  this.useDefaultProfile = false;
  this.userDataDir = this.opts.userDataDir;
}
```

the compiler seems to understand `this.opts.userDataDir` would then be `string|undefined` by the last line there.

Then here in `prepare()` it would just be
```js
this.userDataDir = this.userDataDir || this.makeTmpDir();
```

chrome-launcher.ts
molant: Done

yarn.lock
molant: Done

chrome-launcher.ts
molant: Used your latest comment for the change.

molant:
> would you mind adding a test for this?

Done

molant:
@brendankenny @patrickhulce let me know if you want me to change anything else!

molant:
@brendankenny @patrickhulce ping. Anything else I should do to get this merged?

Thanks!

paulirish:
shipped in 0.9.0

molant:
Thank you all!

glebmachine:
Could you provide any examples of use into documentation?

PR #49, opened 2017-11-08
New: Add flag to use default user profile instead

### LINKED ISSUE #47: How to use default userDataDir on launch?

molant:
Close #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,10 @@ npm install chrome-launcher
   // * Otherwise, a detected Chrome (stable) will be used
   chromePath: string;
 
+  // (optional) Use the default Chrome profile instead of creating a new one
+  // Default: false
+  forceDefaultProfile: boolean;
+
   // (optional) Chrome profile path to use
   // By default, a fresh Chrome profile will be created
   userDataDir: string;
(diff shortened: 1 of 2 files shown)


molant:
This PR address the name discussed in https://github.com/GoogleChrome/chrome-launcher/pull/48#discussion_r149540778 and nothing else.
Other changes are addressed in #48. Only one of this 2 PRs should be merged.

molant:
Closing this PR per the conversation in https://github.com/GoogleChrome/chrome-launcher/pull/48#pullrequestreview-75863412

We are **not** going to add a new property.

PR #51, opened 2017-11-09
fix: support alpine linux by retrying grep with -r

AkshayIyer12:
Chrome-finder will fall back to using lowercase R on platforms that don't support uppercase R.

fixes #46

PR #60, opened 2017-11-30
Improved test of supported grep options

guppie70:
Hi,

(this is my first pull request, so I hope this is the right process/format to help)
I found that running chrome-launcher in an Alpine UNIX based docker image throws errors regarding the grep options used when you try to find the Chrome executable.
Below I have tried to fix that by first running a test grep command over a file that is always present in Linux /etc/hosts
If that throws an error then I use the lower case -r option.
For me this is working now.
Had to submit my delta's so many times as I was debugging the docker container...
Must be a better way to do this :-)
Hope this helps!
Regards,

Johan.


paulirish:
hey @guppie70 

well, first the bad news: we had another two pull requests to fix this issue and we went with them.
here they are: https://github.com/GoogleChrome/chrome-launcher/pull/51 https://github.com/GoogleChrome/chrome-launcher/pull/61

---------

but to give you feedback on your pull request...


thanks very much for contributing! looks like you got the development environment all set up and running. nice stuff. :)

you can see the diff of your PR here: https://github.com/GoogleChrome/chrome-launcher/pull/60/files

As you'll notice there are a bunch of files added.. notably `.d.ts` and `.js` files. Those aren't intended to be committed to the repo, so i suppose they accidentally snuck in via a `git add`.

In your PR you basically test out `grep -ER` first to see if it'll succeed and then choose the correct options based on the result. The PR that we merged also does a try/catch but just tries the real command... if fails, it attempts with the other flag.  see https://github.com/GoogleChrome/chrome-launcher/pull/61/files


And typically you probably would strip out the debugRoutine stuff before submitting the PR, unless it's pretty wild and those messages would assist the reviewer as they try it out. In this case you probably could have omitted them. :)

anyhow.. thanks for the pull request and appreciate your contribution! hopefully there's another opportunity for you to jump in. cheers.  😁

PR #64, opened 2017-12-03
Make chrome-launcher root package node 4 compat.

samccone:
###  Fixes 

- change tsconfig to target es5
- change tsconfig to include es2015 lib polyfills
- change set iteration to use .forEach since this support is not present
  in es5
- make manual-chrome-launcher not use es2015 import syntax

REF #8

### DIFF
--- a/chrome-launcher.ts
+++ b/chrome-launcher.ts
@@ -52,9 +52,12 @@ export interface ModuleOverrides {
 }
 
 const sigintListener = async () => {
-  for (const instance of instances) {
+  // Since sets are not iterable below es2015 we need
+  // too `.forEach` iterate over the set.
+  // https://github.com/Microsoft/TypeScript/issues/6842
+  instances.forEach(async (instance) => {
     await instance.kill();
-  }
+  });
   process.exit(_SIGINT_EXIT_CODE);
 };
 
--- a/manual-chrome-launcher.js
+++ b/manual-chrome-launcher.js
@@ -14,7 +14,7 @@
  */
 
 require('./compiled-check.js')('chrome-launcher.js');
-const {launch} = require('./chrome-launcher');
+const launch = require('./chrome-launcher').launch;
 
 const args = process.argv.slice(2);
 let chromeFlags;
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
         "module": "commonjs",
-        "target": "es2016",
+        "target": "es5",
         "declaration": true,
         "noImplicitAny": true,
         "inlineSourceMap": true,
@@ -9,7 +9,11 @@
         "strictNullChecks": true,
         "noImplicitReturns": true,
         "noUnusedLocals": true,
-        "noUnusedParameters": true
+        "noUnusedParameters": true,
+        "lib": [
+            "es2015",
+            "es2016.array.include"
+        ]
     },
     "exclude": [
         "node_modules"


chrome-launcher.ts
```diff
@@ -52,9 +52,12 @@ export interface ModuleOverrides {
 }
 
 const sigintListener = async () => {
-  for (const instance of instances) {
+  // Since sets are not iterable below es2015 we need
+  // too `.forEach` iterate over the set.
+  // https://github.com/Microsoft/TypeScript/issues/6842
+  instances.forEach(async (instance) => {
```
patrickhulce: won't forEach not wait for the instances to be killed though?

seems like we might need a `await Promise.all(instances.map(i => i.kill())` instead

samccone: since we are awaiting it will :)

example :taco: 

https://www.typescriptlang.org/play/#src=const%20arr%20%3D%20%5B1%2C%202%2C%203%2C%204%5D%3B%0D%0A%0D%0Aasync%20function%20doWork(count)%20%7B%0D%0A%20%20%20%20return%20new%20Promise((res%2C%20_)%20%3D%3E%20%7B%0D%0A%20%20%20%20%20%20%20%20setTimeout(()%20%3D%3E%20res()%2C%20count)%3B%0D%0A%20%20%20%20%7D)%3B%0D%0A%7D%0D%0Aasync%20function%20wow()%20%7B%0D%0A%20%20%20%20arr.forEach(async%20(v)%20%3D%3E%20%7B%0D%0A%20%20%20%20%20%20%20%20await%20doWork(v%20*%201000)%3B%0D%0A%20%20%20%20%20%20%20%20console.log('done%20with'%2C%20v)%3B%0D%0A%20%20%20%20%7D)%3B%0D%0A%7D%0D%0A%0D%0A(async%20()%20%3D%3E%20%7B%0D%0A%20%20%20%20await%20wow()%3B%0D%0A%20%20%20%20console.log('all%20done')%3B%0D%0A%7D)()%0D%0A

samccone:
> want to throw node4 into travis targets?

@patrickhulce I will once i make the parent dep of lighthouse logger work. if I enable 4 right now it will be failing for all contributors. 

Thanks for the review!

chrome-launcher.ts
patrickhulce: hm this example seems to prove my point, unless there's something I'm missing?

running that script prints

```
all done
done with 1
done with 2
done with 3
done with 4
```

when we'd want it to be

```
done with 1
done with 2
done with 3
done with 4
all done
```

and the generated code shows that `wow` just returns immediately after the forEach without awaiting the contents of the loop

```js
return __generator(this, function (_a) {
  // this forEach is not being awaited in any way, the return value is pretty much thrown away here
  arr.forEach(function (v) { return __awaiter(_this, void 0, void 0, function () {
      return __generator(this, function (_a) {
          switch (_a.label) {
              case 0: return [4 /*yield*/, doWork(v * 1000)];
              case 1:
                  _a.sent();
                  console.log('done with', v);
                  return [2 /*return*/];
          }
      });
  }); });
  return [2 /*return*/]; // so it just exits synchronously here
});
```

patrickhulce:
> if I enable 4 right now it will be failing for all contributors.

Ah yeah reviewed this before I saw https://github.com/GoogleChrome/lighthouse/issues/3981 sorry 'bout that :)

chrome-launcher.ts
samccone: oh or a third option, we were making different points :P I follow what you are looking for, I will address it.

Thanks

paulirish:
A few things:

* We're discussing dropping support for a node version in Lighthouse (not chrome-launcher, necessarily) when it. Either when it exits Active (after 6mo) or 3 months before that (which is halfway into the overlap). 
   - It does seem reasonable for c-launcher to have a diff support policy than LH, but I don't think it _needs_ to.
* If we do support node4 it would only be for ~6mo.
* Working with headless-chrome and all that is still a pretty young area, so I think it's reasonable to have shorter support policy than a typical popular module.
* The maintenance cost for c-launcher appears to be pretty non-existent, but lighthouse-logger will get a bit uglier. 
* Adding travis coverage for 4 either means longer build times or we stop building on one of 6/7/8.
* (and now we have no customers awaiting node4 support)

so yeah let's skip this this time..

thanks to sam for investing his time, contributing, and making the case for this.

PR #66, opened 2017-12-06
Enable users to pass env vars into spawned chrome.

### LINKED ISSUE #29: Add ability to pass environment variables for launched Chrome process (to set TZ)
I have a use case where I need to launch Chrome with a different "TZ" variable than the one the environment is running under.

While I can do this in my launching Node process, it would make my script easier if I had a way to pass TZ into the launcher.

If you accept env as a launcher option, it would then get passed along into spawn command here: https://github.com/GoogleChrome/chrome-launcher/blob/master/chrome-launcher.ts#L220

Format of env argument could be something like this for example:

env: { 'TZ': 'America/New_York'}

samccone:
Since it can be useful for users to programatically set env vars
for a given instance of chrome, expose the ability for users to pass
in a env dict into the spawn command that is used to launch chrome.

FIXES #29

---

Sidebar I will refactor the tests to DRY up existing implementations with a follow on CL. I chose to omit it from this CL as it increases the scope of the changeset and muddies the intent of this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,9 @@ npm install chrome-launcher
   // (optional) A number of retries, before browser launch considered unsuccessful.
   // Default: 50
   maxConnectionRetries: number;
+
+  // (optional) A dict of environmental key value pairs to pass to the spawned chrome process.
+  envVars: {[key: string]: string};
 };
 ```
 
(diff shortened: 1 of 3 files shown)


paulirish:
The open PR for enable-audio also cleans up the tests btw

PR #67, opened 2017-12-08
Expose launched chrome child process object.

### LINKED ISSUE #7: Maybe add a notification on 'Close' event?
_From @ghost on June 28, 2017 14:23_

Greetings,

So, we're doing some unit testing of an application. One of such test is to launch a Chrome instance, run some JavaScript to close Chrome. After that, we have to test whether or not the Chrome is actually closed.

In current implementation of `chrome-launcher`, it's very hard to do so because there is no callback which will indicate a passive Chrome shutdown.

However, [it seems that](https://nodejs.org/api/child_process.html#child_process_event_exit) the `child_process` module is actually support `exit` event, so maybe it's possible to add a callback to notify the Chrome instance is gone?

_Copied from original issue: GoogleChrome/lighthouse#2609_

samccone:
Some users may want/need access to the underlying child process as
generated from the spawn command, so expose it when launching chrome.

Also add a test to make sure that the interface that we return does not
have undefined values set.

---

This PR is from #40 but cleaned up so that we can land it.

fixes #7

### DIFF
--- a/chrome-launcher.ts
+++ b/chrome-launcher.ts
@@ -13,6 +13,7 @@ import * as chromeFinder from './chrome-finder';
 import {getRandomPort} from './random-port';
 import {DEFAULT_FLAGS} from './flags';
 import {makeTmpDir, defaults, delay, getPlatform, toWinDirFormat} from './utils';
+import {ChildProcess} from 'child_process';
 const log = require('lighthouse-logger');
 const spawn = childProcess.spawn;
 const execSync = childProcess.execSync;
@@ -42,6 +43,7 @@ export interface Options {
 export interface LaunchedChrome {
   pid: number;
   port: number;
+  chromeProcess: ChildProcess;
   kill: () => Promise<{}>;
 }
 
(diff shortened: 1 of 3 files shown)


chrome-launcher.ts
```diff
@@ -94,7 +96,7 @@ class Launcher {
   private requestedPort?: number;
   private connectionPollInterval: number;
   private maxConnectionRetries: number;
-  private chrome?: childProcess.ChildProcess;
+  public chrome?: childProcess.ChildProcess;
```
paulirish: sg. seems like we can also add private/public/public to the props on L105-107?

chrome-launcher.ts
```diff
@@ -79,7 +81,7 @@ async function launch(opts: Options = {}): Promise<LaunchedChrome> {
     return instance.kill();
   };
 
-  return {pid: instance.pid!, port: instance.port!, kill};
+  return {pid: instance.pid!, port: instance.port!, kill, chromeProcess: instance.chrome!};
```
paulirish: let's rename this guy to just `process`.

chrome-launcher.ts
samccone: oh good catch, public is actually not-needed since when omitted it is the default. I will drop it from here.

chrome-launcher.ts
samccone: +1 will do thanks for the review!

PR #69, opened 2017-12-08
folder refactor: ts in /src, js in /dist

paulirish:
Otherwise the folder just gets quite messy. :/

I'm not 100% this keeps the issue from https://github.com/GoogleChrome/lighthouse/pull/2898 fixed, but I believe it should be fine.

cc @samccone @devrelm


patrickhulce:
Hm probably need a major version bump since the old Readme still has quite a few people requiring `chrome-launcher/chrome-launcher` 🤦‍♂️ unless we want to proxy export with a new file in there?

tests prob need updating too by the looks of travis :)

paulirish:
thx!

proxy export added.
tests updated.

paulirish:
@devrelm can you test this out with your TS setup? I just want to make sure i'm not regressing things here.

PR #87, opened 2018-01-16
Fix Travis CI build

westy92:
This is a fix from https://github.com/travis-ci/travis-ci/issues/8836 that worked for me on [a different project](https://github.com/westy92/html-pdf-chrome).


paulirish:
yeah.. this bug. :)  

kind of a pain. and yah this fix worked for us on LH too.

thanks man.

samccone:
nice work and thanks!

PR #93, opened 2018-01-29
Set custom error codes for all errors

paambaati:
This PR introduces error codes ([further reading](https://medium.com/the-node-js-collection/node-js-errors-changes-you-need-to-know-about-dc8c82417f65)).

**Why?**
It is easier to catch specific errors with something like —

```
try {
    ...
} catch (err) {
    if (err.code === 'ERR_CHROME_PATH_NOT_SET') {}
}
```

instead of —

```
try {
    ...
} catch (err) {
    if (err.message.startsWith(`The environment variable CHROME_PATH`)) {}
}
```

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -38,7 +38,7 @@ export interface Options {
   enableExtensions?: boolean;
   connectionPollInterval?: number;
   maxConnectionRetries?: number;
-  envVars?: {[key: string]: string | undefined};
+  envVars?: {[key: string]: string|undefined};
 }
 
 export interface LaunchedChrome {
@@ -101,7 +101,7 @@ class Launcher {
   private rimraf: typeof rimraf;
   private spawn: typeof childProcess.spawn;
   private useDefaultProfile: boolean;
-  private envVars: {[key: string]: string | undefined};
+  private envVars: {[key: string]: string|undefined};
 
   chrome?: childProcess.ChildProcess;
   userDataDir?: string;
@@ -309,7 +309,6 @@ class Launcher {
             });
       };
       poll();
-
     });
   }
 
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -275,7 +275,7 @@ class Launcher {
   }
 
   // resolves when debugger is ready, rejects after 10 polls
-  private waitUntilReady() {
+  public waitUntilReady(): Promise<void> {
     const launcher = this;
 
     return new Promise((resolve, reject) => {
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -12,7 +12,7 @@ import * as rimraf from 'rimraf';
 import * as chromeFinder from './chrome-finder';
 import {getRandomPort} from './random-port';
 import {DEFAULT_FLAGS} from './flags';
-import {makeTmpDir, defaults, delay, getPlatform, toWinDirFormat, ChromeLauncherError} from './utils';
+import {makeTmpDir, defaults, delay, getPlatform, toWinDirFormat, InvalidUserDataDirectoryError, UnsupportedPlatformError, ChromeNotInstalledError} from './utils';
 import {ChildProcess} from 'child_process';
 const log = require('lighthouse-logger');
 const spawn = childProcess.spawn;
@@ -130,8 +130,7 @@ class Launcher {
         this.useDefaultProfile = true;
         this.userDataDir = undefined;
       } else {
-        throw new ChromeLauncherError(
-            'userDataDir must be false or a path', 'ERR_CHROME_INVALID_USER_DATA_DIRECTORY');
+        throw new InvalidUserDataDirectoryError();
       }
     } else {
       this.useDefaultProfile = false;
@@ -170,8 +169,7 @@ class Launcher {
   prepare() {
     const platform = getPlatform() as SupportedPlatforms;
     if (!_SUPPORTED_PLATFORMS.has(platform)) {
-      throw new ChromeLauncherError(
-          `Platform ${platform} is not supported`, 'ERR_CHROME_UNSUPPORTED_PLATFORM');
+      throw new UnsupportedPlatformError();
     }
 
     this.userDataDir = this.userDataDir || this.makeTmpDir();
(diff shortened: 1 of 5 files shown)


samccone:
I like this idea.

Can we use an exported enum for the codes for each error class?

something like...
```typescript
class FooError implements LauncherError {
  code: CODES.FOO_ERROR,
}
```

paambaati:
@samccone Great idea, I've introduced string enums for the error codes, and created new error subclasses.

Some other changes I had to do —

1. Bump up Typescript version, as string enums were introduced only in 2.4. I figured we might as well move to the latest and the greatest version, so it is 2.7.1.
2. Make the `waitUntilReady()` method `public`, as the tests stub it, and TS is now stricter.

samccone:
I will review the PR this weekend.

Thanks for the work @paambaati !

src/chrome-launcher.ts
```diff
@@ -276,7 +275,7 @@ class Launcher {
   }
 
   // resolves when debugger is ready, rejects after 10 polls
-  private waitUntilReady() {
+  public waitUntilReady(): Promise<void> {
```
samccone: no need for public, as this is inferred/default, and there is no need to set the return type of the function as it is also inferred.

src/utils.ts
```diff
@@ -10,14 +10,56 @@ import {execSync} from 'child_process';
 import * as mkdirp from 'mkdirp';
 const isWsl = require('is-wsl');
 
-export function defaults<T>(val: T | undefined, def: T): T {
+export enum LaunchErrorCodes {
+  ERR_LAUNCHER_PATH_NOT_SET = 'ERR_LAUNCHER_PATH_NOT_SET',
```
samccone: any reason why you went with string enums? vs non-string enums?

src/utils.ts
```diff
@@ -10,14 +10,56 @@ import {execSync} from 'child_process';
...
 }
 
 export async function delay(time: number) {
   return new Promise(resolve => setTimeout(resolve, time));
 }
 
+export class LauncherError extends Error {
+  message: string;
+  code: string;
+  constructor(message: string = 'Unexpected error', code?: string) {
```
samccone:
```ts
constructor(public message = 'unexpected error', public code?: string;) {
   super();
   this.stack = new Error().stack;
}
```

src/utils.ts
```diff
@@ -10,14 +10,56 @@ import {execSync} from 'child_process';
...
+  code: string = LaunchErrorCodes.ERR_LAUNCHER_INVALID_USER_DATA_DIRECTORY;
+}
+
+export class UnsupportedPlatformError extends LauncherError {
+  message: string = `Platform ${getPlatform()} is not supported.`;
+  code: string = LaunchErrorCodes.ERR_LAUNCHER_UNSUPPORTED_PLATFORM;
+}
+
+export class ChromeNotInstalledError extends LauncherError {
+  message: string = 'No Chrome installations found.';
```
samccone: : string can be inferred on all of these.

src/chrome-launcher.ts
paambaati: @samccone Fixed in e3c8ab8.

src/utils.ts
paambaati: @samccone Fixed in fc0d4b9.

src/utils.ts
paambaati: @samccone Fixed in fc0d4b9.

src/utils.ts
paambaati: @samccone They are a little easier to debug — see https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

> string enums have the benefit that they "serialize" well. In other words, if you were debugging and had to read the runtime value of a numeric enum, the value is often opaque - it doesn’t convey any useful meaning on its own (though reverse mapping can often help), string enums allow you to give a meaningful and readable value when your code runs, independent of the name of the enum member itself.

paambaati:
@samccone _BUMP!_

src/utils.ts
```diff
@@ -10,14 +10,56 @@ import {execSync} from 'child_process';
...
 }
 
 export async function delay(time: number) {
   return new Promise(resolve => setTimeout(resolve, time));
 }
 
+export class LauncherError extends Error {
+  public message: string;
+  public code: string;
+  constructor(message: string = 'Unexpected error', code?: string) {
```
samccone: my previous comment about using


constructor(public message: string = '...', public code?: string) {
}

still stands

src/utils.ts
```diff
@@ -10,14 +10,56 @@ import {execSync} from 'child_process';
...
+    }
+    this.stack = new Error().stack;
+    return this;
+  }
+}
+
+export class ChromePathNotSetError extends LauncherError {
+  message =
+      'The environment variable CHROME_PATH must be set to executable of a build of Chromium version 54.0 or later.';
+  code: string = LaunchErrorCodes.ERR_LAUNCHER_PATH_NOT_SET;
```
samccone: all of the code: string bits can be dropped as the type can be inferred.

src/utils.ts
paambaati: @samccone Fixed in c6f69c5.

src/utils.ts
paambaati: @samccone Fixed in c6f69c5.

src/utils.ts
```diff
@@ -10,14 +10,54 @@ import {execSync} from 'child_process';
...
 }
 
 export async function delay(time: number) {
   return new Promise(resolve => setTimeout(resolve, time));
 }
 
+export class LauncherError extends Error {
+  constructor(public message: string = 'Unexpected error', public code?: string) {
+    super();
+    this.message = message;
```
samccone: since you now have the public attr on the arguments in the constructor it means that you can delete
lines 31 though 34

here is a good explainer
https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/

paambaati: @samccone Oh wow, I didn't know this. Fixed in 777aa72.

paambaati:
@samccone **BUMP!**

samccone:
Thanks for your work on this!

PR #99, opened 2018-03-27
Fix picking CHROME_PATH priority over other matches

### LINKED ISSUE #95: How can I launch regular Chrome using CHROME_PATH when Chrome Canary is installed?
When I have both regular Chrome and Chrome Canary installed in their default paths on MacOS I can't find a way to force chrome-launcher to choose regular Chrome instead of Canary using the CHROME_PATH environment variable.

Since my CHROME_PATH value is the same as [one of the built-in priorities in `chrome-finder.ts`](https://github.com/GoogleChrome/chrome-launcher/blob/v0.10.2/src/chrome-finder.ts#L54) and [the sort function returns on the first priority match](https://github.com/GoogleChrome/chrome-launcher/blob/v0.10.2/src/chrome-finder.ts#L195) it will never reach my custom priority since [its added to the end of the list of priorities](https://github.com/GoogleChrome/chrome-launcher/blob/v0.10.2/src/chrome-finder.ts#L65).

And since Canary is installed and matches and it has a higher priority than regular Chrome the Canary installation will be chosen.

The only way I find to force regular Chrome is to use the `chromePath` launch option, but I don't want to set that since other team members are on different systems and need `chrome-finder.ts` to do its work, so I need the CHROME_PATH environment variable.

chrome-launcher @ 0.10.2
MacOS

EvilOatmeal:
Fixes #95

### DIFF
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -58,11 +58,11 @@ export function darwin() {
   ];
 
   if (process.env.LIGHTHOUSE_CHROMIUM_PATH) {
-    priorities.push({regex: new RegExp(`${process.env.LIGHTHOUSE_CHROMIUM_PATH}`), weight: 150});
+    priorities.unshift({regex: new RegExp(`${process.env.LIGHTHOUSE_CHROMIUM_PATH}`), weight: 150});
   }
 
   if (process.env.CHROME_PATH) {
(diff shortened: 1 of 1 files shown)


EvilOatmeal:
CLA signed @googlebot 🤖

samccone:
@EvilOatmeal thanks so much for this improvement!

PR #105, opened 2018-04-13
Stricter typing for logLevel

asakusuma:
Reduce possible set of input values to the allowed input values.


yarn-error.log
```diff
@@ -0,0 +1,1630 @@
+Arguments:
```
paulirish: can you remove this file?

asakusuma:
@paulirish ah sorry about that. I've removed the log file and fixed the formatting as well.

PR #112, opened 2018-06-08
fix: reject promise on failed kill

patrickhulce:
if you've tried to launch canary today you may have noticed some strange behavior...
image

even if you ctrl-c, it just keeps on chugging. turns out we never resolve/reject the promise on a failed kill and so our sig int handler sits there forever

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -58,7 +58,8 @@ const sigintListener = async () => {
   for (const instance of instances) {
     try {
       await instance.kill();
-    } catch (err) {}
+    } catch (err) {
+    }
   }
   process.exit(_SIGINT_EXIT_CODE);
 };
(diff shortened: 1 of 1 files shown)


src/chrome-launcher.ts
```diff
@@ -56,7 +56,10 @@ export interface ModuleOverrides {
 
 const sigintListener = async () => {
   for (const instance of instances) {
-    await instance.kill();
+    try {
+      await instance.kill();
+    } catch (err) {
```
samccone: A log here I think would be a good idea instead of silently eating the error.

src/chrome-launcher.ts
```diff
@@ -314,9 +317,10 @@ class Launcher {
   }
 
   kill() {
-    return new Promise(resolve => {
+    return new Promise((resolve, reject) => {
       if (this.chrome) {
         this.chrome.on('close', () => {
+          delete this.chrome;
           this.destroyTmp().then(resolve);
```
samccone: Hmm can you explain why this line moved? Was there a race in waiting for chrome to close?

src/chrome-launcher.ts
patrickhulce: The only reject path is already logging the error, would you prefer it moved up here?

PR #121, opened 2018-09-09
fix(output): ignore stdio channels (stdout, stdin, stderr)

wardpeet:
Add `{ stdio: 'ignore' }` to execFileSync to make sure no output is generated to our output buffers.

fixes https://github.com/GoogleChrome/lighthouse/issues/5795

### DIFF
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -117,7 +117,8 @@ export function linux() {
   ];
   executables.forEach((executable: string) => {
     try {
-      const chromePath = execFileSync('which', [executable], { stdio: 'ignore' }).toString().split(newLineRegex)[0];
+      const chromePath =
+          execFileSync('which', [executable], {stdio: 'ignore'}).toString().split(newLineRegex)[0];
 
       if (canAccess(chromePath)) {
         installations.push(chromePath);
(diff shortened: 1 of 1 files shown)


wardpeet:
@paulirish whenever you have the time 😄

paulirish:
tbh i can't reproduce the bug on mac or my ubuntu, but i don't see why this wouldn't work.

@wardpeet did you repro it?

wardpeet:
yes I did but only in docker so I guess it has to do with docker piping all output to the docker logs command

vpracht:
The modified line tries to parse the output of the executed command. `{ stdio: 'ignore' }` redirects the output to `/dev/null`, completely breaking that.

wardpeet:
Sorry what are you referring to? What does it break?

wardpeet:
I'll create a fix

wardpeet:
yeah I just tested it in docker too and pipe works as well. The reason why I thought it wouldn't was because I thought the default was pipe as stated in the docs but I missed the following line
> Child's stdio configuration. stderr by default will be output to the parent process' stderr unless stdio is specified. Default: 'pipe'.

wardpeet:
@vpracht fix is up https://github.com/GoogleChrome/chrome-launcher/pull/125

paulirish:
fixed in chrome-launcher@0.10.5

Issue #120, opened 2018-08-24
Killing launched chrome instance in mocha afterEach


mfestrada:
Hi there, 

I have some integration-level tests of a web app that I need to use a headless chrome instance for, and I'm using mocha as a testing framework.  I'd like to use this package for opening/closing the browser in each tests' setup/teardown, but I can't reconcile how to handle the teardown of the browser.  

Say that I have some mocha tests that look like:

```javascript
const chromeLauncher = require('chrome-launcher');

describe('Foo', function() {
  beforeEach(async function() {
    // this part is obviously how I'd launch a browser instance for each test
    chromeLauncher.launch({
      startingUrl: 'https://example.com',
      chromeFlags: ['--headless', '--disable-gpu']
    });
  });

  afterEach(async function() {
    // not sure how to kill the browser I've launched here
  });

  it('Does bar', async function() {
    // testing code that uses chrome
  });
});

```

I don't know how I'm supposed to kill the process in an ```afterEach``` teardown.  I was wondering if someone could point me in the right direction for what I'm trying to do?  Thanks in advance.

patrickhulce:
You'll want to hang on to the chrome instance that's returned

```js
describe('Foo', function() {
  let chrome

  beforeEach(async function() {
    // this part is obviously how I'd launch a browser instance for each test
    chrome = await chromeLauncher.launch({
      startingUrl: 'https://example.com',
      chromeFlags: ['--headless', '--disable-gpu']
    });
  });

  afterEach(async function() {
    // not sure how to kill the browser I've launched here
    await chrome.kill()
  });

  it('Does bar', async function() {
    // testing code that uses chrome
  });
});

patrickhulce:
Also @mfestrada you might want to look into https://github.com/GoogleChrome/puppeteer if you haven't already. It's got a lot of examples and great API to work with chrome for testing.

mfestrada:
@patrickhulce Thank you for your help!!  Sorry to open an issue for something so simple. I've looked into puppeteer, but I had to opt for chrome-remote-interface for some even-lower level capabilities.  All of this headless work you folks are doing is awesome, cheers to you all!

PR #70, opened 2017-12-09
Add option to ignore default flags. BREAKING

samccone:
Originally from
https://github.com/GoogleChrome/chrome-launcher/pull/38

fixes https://github.com/GoogleChrome/chrome-launcher/issues/28

---
high level

⚠️  Breaking API change
* drop enableExtensions boolean

👁  CL specific changes
* add an option to ignore all non-required flags
* refactor tests to use a common spawn helper

As always thank you for reviewing.

PR #124, opened 2018-09-17
remove enableExtensions option. add ignoreDefaultFlags option, add defaultFlags() method

paulirish:
Originally written in #70. Landed and then later reverted in #120 as its a breaking change that necessitates a major release.

(This PR won't be passing due to the typechecking and correct export of defaultFlags)


SirMorfield:
Is there any progress? I would love to see this feature implemented. Without it this package is completely useless to me.

paulirish:
shipped in chrome-launcher@0.11.0

PR #129, opened 2018-10-08
silence stderr on grep commands

### LINKED ISSUE #100: grep fallback logic is noisy
```
grep: unrecognized option: R
BusyBox v1.26.2 (2017-11-23 08:40:54 GMT) multi-call binary.

Usage: grep [-HhnlLoqvsriwFE] [-m N] [-A/B/C N] PATTERN/-e PATTERN.../-f FILE [FILE]...

Search for PATTERN in FILEs (or stdin)

        -H      Add 'filename:' prefix
        -h      Do not add 'filename:' prefix
        -n      Add 'line_no:' prefix
        -l      Show only names of files that match
        -L      Show only names of files that don't match
        -c      Show only count of matching lines
        -o      Show only the matching part of line
        -q      Quiet. Return 0 if PATTERN is found, 1 otherwise
        -v      Select non-matching lines
        -s      Suppress open and read errors
        -r      Recurse
        -i      Ignore case
        -w      Match whole words only
        -x      Match whole lines only
        -F      PATTERN is a literal (not regexp)
        -E      PATTERN is an extended regexp
        -m N    Match up to N times per file
        -A N    Print N lines of trailing context
        -B N    Print N lines of leading context
        -C N    Same as '-A N -B N'
        -e PTRN Pattern to match
        -f FILE Read pattern from file
```

We should silence logging from the grep command so that our fallback to `-r` happens without spamming to the console.

paulirish:
fixes #100

### DIFF
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -233,9 +233,9 @@ function findChromeExecutables(folder: string): Array<string> {
     // Some systems do not support grep -R so fallback to -r.
     // See https://github.com/GoogleChrome/chrome-launcher/issues/46 for more context.
     try {
-      execPaths = execSync(`grep -ER "${chromeExecRegex}" ${folder} | awk -F '=' '{print $2}'`);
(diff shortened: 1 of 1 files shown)


paulirish:
thx #138

PR #134, opened 2018-12-06
Adding an option to enable audio, disabling global flag --mute-audio

randrade23:
Related: #28 #38 #70 #122 #124

This adds an option to .launch to allow enabling audio.

While it has been argued in the mentioned issues that audio is often disabled during automated testing, there are cases in which it is important to be able to play sounds, especially when we are doing tests that are not entirely automated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,10 @@ npm install chrome-launcher
   // Default: false
   enableExtensions: boolean;
 
+  // (optional) Enable audio playback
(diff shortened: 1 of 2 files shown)


paulirish:
#124 has now landed. I've added an example in the readme for how to handle this.
https://github.com/GoogleChrome/chrome-launcher#launching-with-support-for-extensions-and-audio

PR #139, opened 2019-04-18
Removing a dead flag.

asteinha:
See https://chromium.googlesource.com/chromium/src.git/+/ba49d00fddeb39c0816090dad7d359df0dce9a1a


paulirish:
thank yoU!

PR #140, opened 2019-04-24
Document that --enable-automation disables automatic page reloads

flotwig:
Was having trouble with Chrome not automatically retrying requests until I realized that `--enable-automation` breaks that behavior, which should be documented here.


flotwig:
CLA signed

PR #142, opened 2019-04-24
fix: move unneeded ts types to devDeps

### LINKED ISSUE #127: move @types/* packages to devDependencies
If they are in dependencies, it enforces packages that depends on chrome-launcher to install the packages with same versions (@types/core-js, @types/rimraf, @types/mkdirp, @types/node and their dependencies).

So please add them as devDependencies.

patrickhulce:
closes #127

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,9 @@
     "type-check": "tsc --allowJs --checkJs --noEmit --target es2016 *.js"
   },
   "devDependencies": {
+    "@types/mkdirp": "^0.3.29",
     "@types/mocha": "^2.2.41",
+    "@types/rimraf": "^0.0.28",
     "@types/sinon": "^2.3.1",
     "clang-format": "^1.0.50",
     "mocha": "^3.2.0",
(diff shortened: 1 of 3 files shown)


src/chrome-launcher.ts
```diff
@@ -50,7 +52,7 @@ export interface LaunchedChrome {
 
 export interface ModuleOverrides {
   fs?: typeof fs;
-  rimraf?: typeof rimraf;
+  rimraf?: RimrafModule;
```
brendankenny: why'd we need this?

package.json
```diff
@@ -21,10 +23,7 @@
     "typescript": "2.7.1"
   },
   "dependencies": {
-    "@types/core-js": "^0.9.41",
-    "@types/mkdirp": "^0.3.29",
-    "@types/node": "^9.3.0",
-    "@types/rimraf": "^0.0.28",
+    "@types/node": "*",
```
brendankenny: 🎉 🎉

src/chrome-launcher.ts
patrickhulce: well if we want our users to not need `@types/rimraf` then we can't do `import 'rimraf'` in our `d.ts` files.

because we export this type here, we need to replace this. it's just one function so it seems pretty easy to type ourselves

brendankenny: oh, lol, just noticed the interface. Oh `chrome-launcher`...

PR #149, opened 2019-06-20
Fix MacOS Catalina Regex issue

### LINKED ISSUE #146: Support for macOS Catalina
chrome-finder.js has problems to find the path of the installed Chrome app.

It seems the output of lsregister has changed.

macOS 10.15 Beta (19A471t)

ajoy39:
Mac OS Catalina appends a hex code to the end of the path output by LSRegisters, the regex that finds all the chrome executables looked for a newline immediately following the end of the .app. 

Now the regex finds the paths again, and hex code is trimmed off before being tested for file access and added to the array

Fixes #146

### DIFF
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -33,7 +33,7 @@ export function darwin() {
 
   execSync(
       `${LSREGISTER} -dump` +
-      ' | grep -i \'google chrome\\( canary\\)\\?.app.*$\'' +
+      ' | grep -i \'google chrome\\( canary\\)\\?.app .*$\'' +
       ' | awk \'{$1=""; print $0}\'')
       .toString()
       .split(newLineRegex)
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -33,13 +33,13 @@ export function darwin() {
 
   execSync(
       `${LSREGISTER} -dump` +
-      ' | grep -i \'google chrome\\( canary\\)\\?.app$\'' +
+      ' | grep -i \'google chrome\\( canary\\)\\?.app.*$\'' +
       ' | awk \'{$1=""; print $0}\'')
       .toString()
       .split(newLineRegex)
       .forEach((inst: string) => {
         suffixes.forEach(suffix => {
-          const execPath = path.join(inst.trim(), suffix);
+          const execPath = path.join(inst.substring(0, inst.indexOf('.app') + 4).trim(), suffix);
           if (canAccess(execPath)) {
             installations.push(execPath);
           }
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -33,14 +33,14 @@ export function darwin() {
 
   execSync(
       `${LSREGISTER} -dump` +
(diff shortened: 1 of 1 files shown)


src/chrome-finder.ts
```diff
@@ -33,13 +33,13 @@ export function darwin() {
 
   execSync(
       `${LSREGISTER} -dump` +
-      ' | grep -i \'google chrome\\( canary\\)\\?.app$\'' +
+      ' | grep -i \'google chrome\\( canary\\)\\?.app.*$\'' +
```
patrickhulce: that's a bit aggressive (it matches ~15 lines locally for me) 

is there any narrower search we could put on it? a particular length limit perhaps or additional filtering on the JS side?

ajoy39: I added a space between the .app and the wildcard to prevent the regex from returning subdirectories, that took me from 4 results down to 2. One of them is my actual correct chrome install and the other is a copy of chrome in my trash. I could limit the regex to the applications folder without too much trouble, but that could break the command for anyone who has it installed somewhere else. I'm not that great with regular expressions so I may need some help removing the "trashed" version

src/chrome-finder.ts
```diff
@@ -33,13 +33,13 @@ export function darwin() {
 
   execSync(
       `${LSREGISTER} -dump` +
-      ' | grep -i \'google chrome\\( canary\\)\\?.app$\'' +
+      ' | grep -i \'google chrome\\( canary\\)\\?.app .*$\'' +
```
patrickhulce: now this matches nothing on pre-catalina versions like mine :)

how about let's just remove the `$` business completely (leave as `?.app'`) and do the rest of our filtering in JS where it's a little easier to reason about?

ajoy39: You are absolutely right, I didn't check that one on the Mojave mac mini before I pushed it up. I reverted that change, and put a check to pushing to the installations array to make sure items are unique. Between that and the canAccess() function that will ensure that only valid, unique chrome apps are found and added to the array.

ajoy39:
Sure thing, here's the chrome related stuff including entries for Chrome Helper and the Chrome.app that was in my trash. 

[log.txt](https://github.com/GoogleChrome/chrome-launcher/files/3312306/log.txt)

src/chrome-finder.ts
Alhadis: @ajoy39 Just an FYI, `.*$` is redundant in most cases, because it matches everything (including nothing) up to the end of the string. It would have been better to remove `$` altogether and have `/google chrome(canary )?\.app/i` make no assumptions about what characters follow it.

I amended this as part of GoogleChrome/chrome-launcher#197, and felt I should let you know. :-)

ajoy39: @Alhadis oh, thanks for letting me know! I’m not great with regex so any help is appreciated!

ajoy39: @Alhadis oh, thanks for letting me know! I’m not great with regex so any help is appreciated!

PR #159, opened 2019-07-09
tests: drop support for node 9. continue supporting node 8 LTS

paulirish:
as mentioned here https://github.com/GoogleChrome/chrome-launcher/pull/158#pullrequestreview-259793476


paulirish:
TBR=patrickhulce

PR #162, opened 2019-07-18
fix: Prevent mutation of default flags

Gudahtt:
The default Chrome flags were being mutated by the `flags` method. This led to changes in the default flags over time, which presumably was not intended.

The default flags are now shallowly cloned to prevent mutation. The TypeScript type has been set to `ReadonlyArray` to prevent this from happening in the future.

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -144,7 +144,7 @@ class Launcher {
   }
 
   private get flags() {
-    const flags = this.ignoreDefaultFlags ? [] : DEFAULT_FLAGS;
+    const flags = this.ignoreDefaultFlags ? [] : DEFAULT_FLAGS.slice();
     flags.push(`--remote-debugging-port=${this.port}`);
 
     if (getPlatform() === 'linux') {
@@ -164,7 +164,7 @@ class Launcher {
   }
 
   static defaultFlags() {
-    return DEFAULT_FLAGS;
+    return DEFAULT_FLAGS.slice();
   }
 
   // Wrapper function to enable easy testing.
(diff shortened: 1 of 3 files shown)


src/chrome-launcher.ts
```diff
@@ -144,7 +144,7 @@ class Launcher {
   }
 
   private get flags() {
-    const flags = this.ignoreDefaultFlags ? [] : DEFAULT_FLAGS;
+    const flags = this.ignoreDefaultFlags ? [] : DEFAULT_FLAGS.slice();
```
patrickhulce: shouldn't we be `.slice`'ing the `defaultFlags` return on line 167?

Gudahtt: I did consider doing that - it's probably a good idea. It would certainly be safer.

I held off initially because it's not really necessary for any projects using TypeScript, and avoiding it would mean avoiding duplicating the array. But in retrospect, that's such a tiny micro-optimization that it's hardly worth considering. I'll update it.

Gudahtt:
Sure, good idea. I'll add one shortly.

test/chrome-launcher-test.ts
```diff
@@ -113,6 +113,20 @@ describe('Launcher', () => {
     assert.deepStrictEqual(flags, DEFAULT_FLAGS);
   });
 
+  it('does not allow mutating default flags', async () => {
+    const flags = Launcher.defaultFlags();
+    flags.push('--new-flag')
```
patrickhulce:
```suggestion
    flags.push('--new-flag');
```

AVVS:
Hey, could you cut a release with this fix?

patrickhulce:
@AVVS done in 0.11.2 👍

Issue #88, opened 2018-01-16
remove --disable-translate flag


patrickhulce:
--disable-translate hasn't been a thing for a while now, we can get rid of it from our default flags

rokobuljan:
When opening in kiosk mode it's pretty annoying to see translation suggestions. Why is the flag been removed?

paulirish:
FWIW it looks like Chromium may not have translate enabled, so using that rather than Chrome may be a workaround. I don't know of a decent solution currently, aside from that. 

I agree it sucks the flag was removed without a decent alternative.

PR #167, opened 2019-08-15
Update the translation flag

benschwarz:
In this PR: 

- [x] Update the disable translation flag

---

I noticed that Lighthouse currently displays the translation UI despite chrome-launcher using the `--disable-translation` flag. 

The `--disable-translation` flag is no longer listed on [Peter Beverloo's command line switches page](https://peter.sh/experiments/chromium-command-line-switches/). 

It seems that to achieve the same result, the `--disable-features=TranslateUI` flag must be used. 

closes https://github.com/GoogleChrome/chrome-launcher/issues/88

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -7,7 +7,7 @@
 
 export const DEFAULT_FLAGS: ReadonlyArray<string> = [
   // Disable built-in Google Translate service
-  '--disable-translate',
+  '--disable-features=TranslateUI',
   // Disable all chrome extensions entirely
   '--disable-extensions',
   // Disable various background network services, including extension updating,


benschwarz:
More info in this [chrome bug comment](https://bugs.chromium.org/p/chromium/issues/detail?id=770776#c11).

src/flags.ts
```diff
@@ -7,7 +7,7 @@
 
 export const DEFAULT_FLAGS: ReadonlyArray<string> = [
   // Disable built-in Google Translate service
-  '--disable-translate',
+  '--disable-features=TranslateUI',
```
patrickhulce: is this going to be a problem for folks that are passing their own `--disable-features` if it's supposed to be a csv? should we be combining them?

paulirish: > should we be combining them?

probably yes. doesn't seem a high priority but yah

Issue #169, opened 2019-09-10
How to make sure a tab in background will keep running?


VikramTiwari:
Scenario:
I am launching chrome with following flags:
```
      '--disable-background-timer-throttling',
      '--enable-automation',
      '--disable-renderer-backgrounding',
      '--disable-backgrounding-occluded-windows',
      '--disable-ipc-flooding-protection',
```

And then utilizing puppeteer to do some operations in a tab in the background while I keep focus on a separate tab.

Expectation: The tasks should happen at normal pace in the background tab
Reality: The tasks in the background tab are throttled

Please let me know if there are other details that I can provide. Thanks!

patrickhulce:
Thanks for filing!

Q: You're using headful Chrome, correct? If so, why not headless?

Q: If you're using puppeteer, what led you to use `chrome-launcher` over `puppeteer.launch`?

I'm not aware of a way to completely coerce a headful tab to behave like it's the foreground tab, but maybe there are other workarounds for your use case.

VikramTiwari:
Why headful Chrome?
I am trying to run a script where I might have to do some manual interactions that cannot be automated. So if chrome is running in headless then I won't be able to do those interactions. Though I would love to have a way where jobs are running in background and when a specific function is executed, the window becomes headful. The closest I can get to it is by running the task in background tab and bringing the tab to front if I encounter my manual interaction cases.

Why `chrome-launcher` with `puppeteer`?
Chrome launcher helps me launch chrome with my current user's data directory as well as it gives me the ability to use devtools protocol for certain functions which are not doable via puppeteer. Also, this way I get to simulate the script environment closest to my actual environment.

Other ways to make tab behave like foreground tab:
I have seen examples where we could run a non-audible audio file in background in a loop. But that practice doesn't seem great. Are there any other ways you are familiar with @patrickhulce?

patrickhulce:
I'm not aware of any better ways than the flags you've documented and the autoplay audio and video hacks unfortunately :/

VikramTiwari:
hmm.. that's not great. In my understanding it's done for performance reasons. So, I was really hoping that this would be just a flag. Is there some reason why it's not? Should I open a feature request on crbug?

cc: @paulirish Are there any other methods that we might be missing here?

paulirish:
patrick and i know pretty much the same amount on this topic :)

but yeah i think a crbug makes sense if we can prove this is happening. a test page that prints out timer overage would help a lot.

[...]

PR #170, opened 2019-09-10
flags: add new --disable flags to reduce noise and disable backgrounding

paulirish:
described mostly here: https://blog.bigbinary.com/2018/08/15/debugging-failing-tests-in-background-tab-in-puppeteer.html

also relevant: https://github.com/GoogleChrome/puppeteer/blame/master/lib/Launcher.js

test page for occlusion and timer drift: https://output.jsbin.com/goduxox/1/quiet

ref #169

### DIFF
--- a/docs/chrome-flags-for-tools.md
+++ b/docs/chrome-flags-for-tools.md
@@ -99,7 +99,10 @@ These flags are being used in various tools. They also just need to be documente
 
 --disable-backgrounding-occluded-windows
 --disable-ipc-flooding-protection # https://crrev.com/604305
---disable-renderer-backgrounding
+
+--disable-renderer-backgrounding # This disables non-foreground tabs from getting a lower process priority
+                                 # This doesn't (on its own) affect timers or painting behavior.
+                                 # https://github.com/karma-runner/karma-chrome-launcher/issues/123
 
 --remote-debugging-pipe # more secure than using protocol over a websocket
 --enable-logging=stderr # Logging behavior slightly more appropriate for a server-type process.
--- a/docs/chrome-flags-for-tools.md
+++ b/docs/chrome-flags-for-tools.md
@@ -31,7 +31,7 @@ Skip first run wizards
 Disable timers being throttled in background pages/tabs
 
 ### `--disable-client-side-phishing-detection`
-Disables client-side phishing detection. 
+Disables client-side phishing detection.
 
 ### `--disable-popup-blocking`
 Disable popup blocking.  `--block-new-web-contents` is the strict version of this.
@@ -94,6 +94,10 @@ These flags are being used in various tools. They also just need to be documente
 --disable-features=site-per-process # Disables OOPIF. https://www.chromium.org/Home/chromium-security/site-isolation
 --disable-hang-monitor
 
+--disable-backgrounding-occluded-windows
+--disable-ipc-flooding-protection # https://crrev.com/604305
+
+
 --remote-debugging-pipe # more secure than using protocol over a websocket
 --enable-logging=stderr # Logging behavior slightly more appropriate for a server-type process.
 --log-level=0 # 0 means INFO and higher.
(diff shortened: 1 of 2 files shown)


docs/chrome-flags-for-tools.md
```diff
@@ -94,6 +94,10 @@ These flags are being used in various tools. They also just need to be documente
 --disable-features=site-per-process # Disables OOPIF. https://www.chromium.org/Home/chromium-security/site-isolation
 --disable-hang-monitor
 
+--disable-backgrounding-occluded-windows
```
patrickhulce: seems like it shouldn't be in both lists?

docs/chrome-flags-for-tools.md
```diff
@@ -94,6 +94,10 @@ These flags are being used in various tools. They also just need to be documente
 --disable-features=site-per-process # Disables OOPIF. https://www.chromium.org/Home/chromium-security/site-isolation
```
patrickhulce: add --enable-automation to this list?

brendankenny:
> fixes #169

seems like the problem wasn't fixed by adding those flags in that issue? I'm confused :)

src/flags.ts
```diff
@@ -10,6 +10,8 @@ export const DEFAULT_FLAGS: ReadonlyArray<string> = [
   '--disable-features=TranslateUI',
   // Disable all chrome extensions entirely
   '--disable-extensions',
+  // Disable some extensions that aren't affected by --disable-extensions
```
brendankenny: should probably remove `entirely` from L11 then :)

docs/chrome-flags-for-tools.md
paulirish: in fact we have the best docs for that particular flag. ;) https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md#--enable-automation

docs/chrome-flags-for-tools.md
paulirish: the md is just for other tool developers and hackers like vikram who are digging into this. ;)

paulirish:
> Side effect of making GoogleChrome/lighthouse#9650 difficult to test?

yup. will have to relaunch without that flag in order to test. oh boy.

docs/chrome-flags-for-tools.md
patrickhulce: oh I see now :)

docs/chrome-flags-for-tools.md
patrickhulce: oh haha nice, I wish GH had a ctrl+f for the whole file easily available without having to manually expand the undiff'd parts of it

VikramTiwari:
Just wanted to chime in and [confirm](https://github.com/GoogleChrome/chrome-launcher/pull/170#issuecomment-530137047) that even after adding these flags the pages in the backgrounded tab were throttled. 

Though, I am not sure if it was just the render, timer or actual node code running through puppeteer that was throttled. I can do more digging if it's unknown/helpful.

paulirish:
@VikramTiwari thanks. that's useful. do you have a minimal test page we can experiment with?

also fwiw here are all the chromium flags: https://peter.sh/experiments/chromium-command-line-switches/

VikramTiwari:
I was trying this on my [LinkedIn profile page](https://www.linkedin.com/in/vikramtiwari/) but you should be able to do it on any profile. Actions were basic copying and pasting.

Minimal script would:
- Load the profile page in a new page
- Use puppeteer's `page.bringToFront()` to bring some other page to front (easiest is `pages[0]`)
- perform copy operations on the older page

Best would be to add multiple copy operations so that you can see the throttle happening. I will get started on creating this into a gist that we can act as proof.

VikramTiwari:
Seems like I am unable to reproduce the issue in my [POC](https://gist.github.com/VikramTiwari/77d0b81e605b0af161e0021e130d34d3) when I use the required flags. I can recreate it in my actual project though. So most probably it's on my side. Will update on the issue if I find something.

VikramTiwari:
Okay! I can reproduce the issue now. Seems like it's just `mousemove`s which cause the slow down to happen. I have updated the POC with random mousemoves between copy events. If you comment [randomMouseMoves](https://gist.github.com/VikramTiwari/77d0b81e605b0af161e0021e130d34d3#file-index-js-L45) the execution runs as normal expected speed.

paulirish:
https://output.jsbin.com/goduxox/1/quiet shows that with these flags....

- timers are no longer throttled when occluded or in a background tab
- the page visibility api/events are unaffected (it still works)
- renderer painting still works when the window is occluded
- however renderer painting doesn't work when the tab is backgrounded.

So this won't fix cli NO_FCP issues entirely but will help.

paulirish:
I'm ready to land this (as soon at CI is done)

PR #175, opened 2019-10-18
deps: update assorted dependencies

brendankenny:
makes github UI warnings go away (all in transitive devDependencies)

Slightly more than usual, though: two of the bad dependencies were deep in `nyc`'s transitive deps and didn't trivially update to non-vulnerable versions. Rather than update the version of nyc, make sure everything is working as before, etc, I just removed nyc :)

I don't think I've ever used coverage on `chrome-launcher` since it split from the `lighthouse` repo and we don't upload the coverage reports anywhere...any reason to keep?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,7 @@ cache:
 install:
   - yarn
 before_script:
-  - export DISPLAY=:99.0
   - export CHROME_PATH="$(pwd)/chrome-linux/chrome"
-  - sh -e /etc/init.d/xvfb start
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -4,7 +4,6 @@ matrix:
     - node_js: "8"
     - node_js: "10"
     - node_js: "12"
-dist: trusty
```
brendankenny: I figured while I was in here removing coverage might as well update

PR #177, opened 2019-10-25
feat: expose public interface for locating Chrome installations

### LINKED ISSUE #176: request: expose public interface for locating chrome path
The business logic used by `chrome-launcher` to locate the chrome binary is robust and well maintained. Unfortunately, it isn't exposed as part of the package's public interface.

Given `chrome-launcher` is effectively the authority for the best / most correct way to launch chrome, IMO, it makes sense to allow public access to this business logic instead of inadvertently encouraging others to attempt to reproduce its behavior.

In short, I'd like to expose a public interface for locating the chrome binary --essentially encapsulating [this logic from `src/chrome-launcher.ts`](https://github.com/GoogleChrome/chrome-launcher/blob/e67a10df/src/chrome-launcher.ts#L208-L213) in a discrete method and exporting it.

How do y'all feel about that?

knksmith57:
Add `Launcher.getInstallations()` public static method that returns an
array of paths to available Chrome binaries. Also, refactor
`Launcher#launch()` to use it.

`Launcher.getInstallations()` uses `chrome-finder` under the hood, so all
previous resolution logic is preserved:
- if `CHROME_PATH` env variable is set, it will be included
- if a Chrome Canary is detected, it will be included
- if a Chrome (stable) is detected, it will be included

Resolves #176

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -167,6 +167,10 @@ class Launcher {
     return DEFAULT_FLAGS.slice();
   }
 
+  static getInstallations() {
+    return chromeFinder[getPlatform() as SupportedPlatforms]();
+  }
+
   // Wrapper function to enable easy testing.
   makeTmpDir() {
     return makeTmpDir();
@@ -205,7 +209,7 @@ class Launcher {
       }
     }
     if (this.chromePath === undefined) {
-      const installations = await chromeFinder[getPlatform() as SupportedPlatforms]();
+      const installations = Launcher.getInstallations();
       if (installations.length === 0) {
         throw new ChromeNotInstalledError();
       }
--- a/test/chrome-launcher-test.ts
+++ b/test/chrome-launcher-test.ts
@@ -133,6 +133,12 @@ describe('Launcher', () => {
     assert.ok(!chromeFlags.includes('--disable-extensions'));
   });
 
+  it('searches for available installations', async () => {
+    const installations = Launcher.getInstallations();
+    assert.ok(Array.isArray(installations));
+    assert.ok(installations.length >= 1);
+  });
+
   it('removes --user-data-dir if userDataDir is false', async () => {
     const spawnStub = await launchChromeWithOpts();
     const chromeFlags = spawnStub.getCall(0).args[1] as string[];
--- a/README.md
+++ b/README.md
@@ -106,6 +106,12 @@ Returns an `Array<string>` of the default [flags](docs/chrome-flags-for-tools.md
 
 Note: This array will exclude the following flags: `--remote-debugging-port` `--disable-setuid-sandbox` `--user-data-dir`.
 
+### `ChromeLauncher.getInstallations()`
+
+Returns an `Array<string>` of paths to available Chrome installations. When `chromePath` is not provided to `.launch()`, the first installation returned from this method is used instead.
+
+Note: This method performs synchronous I/O operations.
+
 ## Examples
 
 #### Launching chrome:


knksmith57:
@googlebot I signed it!

src/chrome-launcher.ts
```diff
@@ -167,6 +167,10 @@ class Launcher {
     return DEFAULT_FLAGS.slice();
   }
 
+  static getInstallations() {
+    return chromeFinder[getPlatform() as SupportedPlatforms]();
```
knksmith57: note: `await` was previously used here; however, afaict, `chromeFinder[]()` methods intentionally use blocking I/O and synchronous interfaces --I opted to implement `getInstallations()` as _non_-async to avoid any ambiguity to the caller as to whether or not the method was blocking.

happy to change this if y'all feel differently 😄

src/chrome-launcher.ts
```diff
@@ -205,7 +209,7 @@ class Launcher {
       }
     }
     if (this.chromePath === undefined) {
-      const installations = await chromeFinder[getPlatform() as SupportedPlatforms]();
+      const installations = (<typeof Launcher>this.constructor).getInstallations();
```
patrickhulce: can't we just call `Launcher.getInstallations()`? :)

knksmith57: in case it's not obvious, because `Launcher` is publicly exported, rather than reference `Launcher.getInstallations()`, I opted to use late static binding here so that consumers that wish to override `getInstallations()` can.

test/chrome-launcher-test.ts
```diff
@@ -133,6 +133,11 @@ describe('Launcher', () => {
     assert.ok(!chromeFlags.includes('--disable-extensions'));
   });
 
+  it('searches for available installations', async () => {
+    const installations = Launcher.getInstallations();
```
patrickhulce: assert that we found one too? :)

src/chrome-launcher.ts
patrickhulce: I'm not sure what value that really has. We already provide `chromePath` as an option and we'd never do anything with installations past the first one.

test/chrome-launcher-test.ts
```diff
@@ -133,6 +133,11 @@ describe('Launcher', () => {
...
+    assert.ok(Array.isArray(installations));
```
knksmith57: this test case is super weak sauce. very much open to, and would appreciate, suggestions for how to better test this feature.

as it stands, this coverage at least ensures that the method exists, is callable, and returns an array --which is better than nothing, but fails to ensure that the appropriate resolution logic is actually used.

I struggled with how to go about testing this method for a few reasons:
1. the project doesn't appear to include any module mocking dependencies and I didn't feel comfortable pulling one in without first reaching consensus that I should
2. all of the heavy lifting is done by chrome-finder, while platform resolution logic is handled by utils; short of asserting that `getPlatform()` and `chromeFinder[]()` are invoked (which likely requires a mocking lib ^^), I'm not sure what else to assert.

src/chrome-launcher.ts
knksmith57: wow you were fast! thanks for hopping in so quickly!

I was proactively typing up a note for why I opted for this but was obvs too slow lol --I'll respond back in the other thread.

knksmith57: that's a super fair and valid point.

I suppose my inclination is to lean towards ensuring that public class methods can be overridden by consumers of the library --but you've made a great point that, currently, the only call-site that uses this already supports being overridden via another interface.

I'll replace with `Launcher.getInstallations()` 😄

test/chrome-launcher-test.ts
knksmith57: I wasn't sure if that was safe given we don't currently require chrome as a dev dependency (that I can tell), so CI might not pass?

src/chrome-launcher.ts
knksmith57: done!

knksmith57: addressed in other thread --closing out :)

test/chrome-launcher-test.ts
knksmith57: OK I added the check --hopefully that works as intended 😆

knksmith57:
OK looks like adding that additional assert worked (I didn't realize we codified the chrome dep in our travis config; I should have looked at that first 😅).

I think I've addressed all feedback --thanks again for the _super_ fast review!

patrickhulce:
yup the rest of our tests would fail if we didn't have chrome available so it's a safe bet :)

patrickhulce:
looks great, thanks again @knksmith57! 🎉 💯

paulirish:
great idea :D

PR #179, opened 2019-11-07
fix: log taskkill error based on logging opts (#178)

### LINKED ISSUE #178: "Process could not be terminated" error is thrown on Windows
First of all, thanks for this awesome library! I use it together with puppeteer-core to avoid chromium installs if Chrome exists on users system, in the library that I maintain: https://github.com/onderceylan/pwa-asset-generator

I'm getting the following error on Windows, after calling `chrome.kill()`. 

```
ERROR: The process with PID 12345 (child process of PID 1234) could not be terminated.
Reason: There is no running instance of the task.
```

I see 2 issues with this;

1. Not sure why but sometimes this error occur and sometimes not. Ideally, I hope the root cause of this can be found and fixed.

2. Second issue is, this error is being logged to the console directly, and logLevel setting is not respected when this error is thrown. Ideally, it should be logged based on logLevel setting.

I hope it's possible to get this issue fixed. If not, I'd love to go for the second option as this issue logs an unnecessary output for my CLI lib at the moment.

onderceylan:
fixes #178

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -332,7 +332,9 @@ class Launcher {
         log.log('ChromeLauncher', `Killing Chrome instance ${this.chrome.pid}`);
         try {
           if (isWindows) {
-            execSync(`taskkill /pid ${this.chrome.pid} /T /F`);
+            // While pipe is the default, stderr also gets printed to process.stderr
+            // if you don't explicitly set `stdio`
+            execSync(`taskkill /pid ${this.chrome.pid} /T /F`, {stdio: 'pipe'});
           } else {
             process.kill(-this.chrome.pid);
           }


onderceylan:
@googlebot I signed it!

src/chrome-launcher.ts
```diff
@@ -332,7 +332,7 @@ class Launcher {
         log.log('ChromeLauncher', `Killing Chrome instance ${this.chrome.pid}`);
         try {
           if (isWindows) {
-            execSync(`taskkill /pid ${this.chrome.pid} /T /F`);
+            execSync(`taskkill /pid ${this.chrome.pid} /T /F`, {stdio: 'pipe'});
```
patrickhulce: add a comment that basically says while pipe is the default, stderr also gets printed to process.stderr if you don't explicitly set it? I could see someone forgetting this and removing it in a driveby someday :)

onderceylan: Sure, feels safer :)

onderceylan:
Hi @patrickhulce, I made the requested changes and rebased the commits. It's ready for re-review.

src/chrome-launcher.ts
```diff
@@ -332,7 +332,9 @@ class Launcher {
         log.log('ChromeLauncher', `Killing Chrome instance ${this.chrome.pid}`);
         try {
           if (isWindows) {
-            execSync(`taskkill /pid ${this.chrome.pid} /T /F`);
+            // While pipe is the default, stderr also gets printed to process.stderr
+            // if you don't explicitly set it
```
patrickhulce:
```suggestion
            // if you don't explicitly set `stdio`
```

onderceylan: Done

PR #182, opened 2019-12-11
flags: disable the default browser check (#181)

### LINKED ISSUE #181: Make the browser not prompt the user to set it as the default browser
### Is this a feature request or a bug?
This is a feature request.

### What is the current behavior?
When starting the browser it checks whether it is the default browser and prompts the user to set it as such when it isn't.

### What is the expected or desired behavior?
The browser should not prompt the user to set it as the default browser.

mariyan-borisov:
It appears that the `--no-default-browser-check` flag is still there and still functioning.
https://github.com/chromium/chromium/blob/cbbe680/chrome/browser/ui/startup/startup_browser_creator_impl.cc#L870-L876

closes #181

### DIFF
--- a/docs/chrome-flags-for-tools.md
+++ b/docs/chrome-flags-for-tools.md
@@ -27,6 +27,9 @@ Disable installation of default apps on first run
 ### `--mute-audio`
 Mute any audio
 
+### `--no-default-browser-check`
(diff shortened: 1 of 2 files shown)


mariyan-borisov:
@googlebot I signed it!

Issue #183, opened 2019-12-31
Option for exit on SIGINT


crislin2046:
This project has saved me a lot of time instead of writing it myself, really grateful for that!!! Thanks 😃 

It would be nice to switch off the `process.exit(...)` behavior that occurs on SIGINT.

What if `chrome-launcher` is embedded by another program that has its own handler on SIGINT? In that case chrome-launcher's handler might kill the node process before the other program's SIGINT handler has completed.

You *can* have async actions in a SIGINT handler. And I don't like the workaround of spawning a new node process just to run chrome launcher. 

I think a switch like, 

```js
   sigintDefaultAction: false, // or
   killProcess: false, // or
   killProcessOnSigint: false, // you get the idea
```

crislin2046:
Closed in #184

crislin2046:
Closed because idea moves the problem around, not solves it.

PR #184, opened 2019-12-31
Add exitOnSIGINT option

crislin2046:
Closes https://github.com/GoogleChrome/chrome-launcher/issues/183

This option lets you control whether chrome-launcher exits the node process when it intercepts a SIGINT signal. 

```ts
// (optional) Exit the Node process on `Ctrl-C`
// Default: true
exitOnSIGINT: boolean;
```

Setting this to `false` is a great idea if you handle `SIGINT` in your code and need to ensure `process.exit` is only called by you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,10 @@ npm install chrome-launcher
   // (optional) Close the Chrome process on `Ctrl-C`
   // Default: true
   handleSIGINT: boolean;
+  
+  // (optional) Exit the Node process on `Ctrl-C`
+  // Default: true
+  exitOnSIGINT: boolean;
 
   // (optional) Explicit path of intended Chrome binary
   // * If this `chromePath` option is defined, it will be used.
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -4,7 +4,6 @@
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
  */
 'use strict';
-
 import * as childProcess from 'child_process';
 import * as fs from 'fs';
 import * as net from 'net';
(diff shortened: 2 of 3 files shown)


patrickhulce:
Thanks very much for the PR @crislin2046! :D

I'm not sure I understand the need for this though. If you're already handling SIGINT yourself, why not just close the chrome instances's there? If we need to `export` the instances set to make this easier, that seems reasonable, but the boolean for fine-grained control of a simple SIGINT handler a bit less so :)

crislin2046:
It sounds like you don't see the need for this, and I get that this fine grained control of a simple SIGINT handler seems like an unreasonable API. I totally understand that.

I think it's a good separation of concerns that chrome launcher kills it's own instances, I don't think I should have to worry about that. 

Also it's good that CL kills instances on SIGINT. Apps and libraries can have their own SIGINT handlers doing their own cleanup, that's a good thing. But when one library calls exit in its handler, it can prevent other SIGINT handlers from completing, which is not intended.

I think enabling the control but keeping the cleanup is good separation of concerns. A library shouldn't be interfering in cleanup processes beside it's own and users of chrome launcher are not responsible for doing the cleanup for chrome launcher. So having an API to control of who calls exit makes sense. That's how I see it. 

Understand if you differ.

Anyway, @patrickhulce , love your work on lighthouse BTW, particularly the dual extension and node nature of it! That's really cool :smile:

patrickhulce:
> Apps and libraries can have their own SIGINT handlers doing their own cleanup, that's a good thing. But when one library calls exit in its handler, it can prevent other SIGINT handlers from completing, which is not intended.

All true and good points, but this approach stops working when handlers need to do some async work. Once we introduce multiple async handlers, someone (i.e. the app itself) needs to be in charge of coordinating them and why `handleSIGINT` is something that can be turned off.

I'm against this option because Chrome Launcher's handler is async and this option only makes sense when you need to do other cleanup work in a SIGINT handler. In that situation, they all need to be coordinated or someone's cleanup is going to go unfinished.

WDYT about just exposing `instances`, then...

A) The app doesn't have any other SIGINT handlers, `handleSIGINT: true` works.
B) The app has other *synchronous* SIGINT handlers, `handleSIGINT: true` works.
C) The app has other *asynchronous* SIGINT handlers, `handleSIGINT: false` + `await Promise.all(instances.map(chrome => chrome.kill().catch(() => {})))` works

crislin2046:
Thanks for the comprehensive answer. I'm afraid I'm against users of c-l doing the cleanup for c-l. Also, I think I see the need for c-l's asynchronous SIGINT handler and why this PR could break that. 

In effect, this PR could cause for c-l the very same problem that c-l could cause for users of c-l. Neither case is ideal, and because this PR is just moving the problem around, not fixing it, I'll close it.

In terms of a fix, I think a simple:

D) if asynchronous cleanup before process.exit, then `handleSIGINT: false` and `await <chrome-launcher>.killAll()` 

will be good. 

Then c-l can handle its own kill implementation, and c-l users can also wait for that to finish, solving the problem, not moving it around. :smile:

PR #186, opened 2020-01-03
Add async killAll

### LINKED ISSUE #185: async killAll function for all launched broswers?
As discussed in #184 there can be conflicts between asynchronous SIGINT handlers registered in different parts of a program, leading to one handler calling process.exit before all work from all handlers is complete. 

If a Promise returning `killAll` function was exposed, this would provide a way to wait for all browsers to close before calling `process.exit`, and could be combined with `handleSIGINT: false` to stop c-l from calling `process.exit`.

Effectively this would give control over when process.exit is called, and also provide a simple async API for closing browsers, without you as a user of c-l needing to care about how that cleanup was done, nor doing it yourself, only needing to care about when you want to do it, then asking c-l to do its own cleanup for you by instructing it to with `killAll`.

crislin2046:
async `killAll` lets you wait to kill all chrome instances created with `launch`

Typically used with `handleSIGINT: false`

This allows you to cleanup Chrome instances without needing to use `handleSIGINT: true` which results in `process.exit` being called on SIGINT and may terminate Node before any other asynchronous cleanup work is done. 

Use `killAll` when you need control over when to call process.exit, and want to wait for Chrome instances to be killed. 

Closes #185


src/chrome-launcher.ts
```diff
@@ -90,6 +85,19 @@ async function launch(opts: Options = {}): Promise<LaunchedChrome> {
   return {pid: instance.pid!, port: instance.port!, kill, process: instance.chrome!};
 }
 
+async function killAll(): Promise<boolean> {
```
patrickhulce: a bit odd IMO to return a boolean here since it's not really actionable to the user, what about returning an array of the errors so they could at least log the messages if they care about them?

src/chrome-launcher.ts
```diff
@@ -90,6 +85,19 @@ async function launch(opts: Options = {}): Promise<LaunchedChrome> {
...
+  let noErrors = true;
+  for (const instance of instances) {
+    try {
+      await instance.kill();
+      instances.delete(instance);
```
patrickhulce: is it intentional to only remove the instance from the set after it's been successfully killed?

on the one hand, it'd be nice for a potential future `killAll` to try again

on the other hand, if it's erroring because it's already dead it's annoying to have this permanently erroring unnecessarily

I think we just need some separate better upkeep of our `instances` set as processes get killed/die on their own buuuuuuuut

either way let's add a comment here about the removal behavior, so future readers know what's up :)

README.md
```diff
@@ -112,6 +112,18 @@ Returns an `Array<string>` of paths to available Chrome installations. When `chr
 
 Note: This method performs synchronous I/O operations.
 
+### `.killAll()`
```
patrickhulce: thanks for the docs! 👍 👍

src/chrome-launcher.ts
crislin2046: it was intentional in order to retry. but good point about the error being caused by already dead instance, I didn't think of that. 

I'll add a comment

src/chrome-launcher.ts
crislin2046: yeah I didn't know if you'd balk at adding complexity so I hedged with a simple errors occurred flag.

I'm not sure the array is the best idea, but it might be. I agree providing more information, if we are going to provide some, does make sense. 

I'll add something.

thanks for the review 😊

crislin2046:
I started having a look at tests but didn't really know where to start. could you please suggest a type of rest to run and I'll see how I go with it

patrickhulce:
oops sorry for the delay @crislin2046 !

> could you please suggest a type of rest to run and I'll see how I go with it

I'm thinking of just adding another entry like https://github.com/GoogleChrome/chrome-launcher/blob/b8c89f84a39c2f4ed4dd547be5dce2d83da54424/test/chrome-launcher-test.ts#L94-L99 except we launch two and invoke `killAll` once :)

crislin2046:
Thanks

README.md
```diff
@@ -112,6 +112,18 @@ Returns an `Array<string>` of paths to available Chrome installations. When `chr
...
+
+Attempts to kill all Chrome instances created with [`.launch([opts])`](#launchopts). Returns a Promise that resolves to `true` if there were no errors, and to `false` if any errors were thrown when killing an instance.
```
patrickhulce:
```suggestion
Attempts to kill all Chrome instances created with [`.launch([opts])`](#launchopts). Returns a Promise that resolves to an array of errors that occurred while killing instances. If all instances were killed successfully, the array will be empty.
```

crislin2046: Thanks for catching!

crislin2046:
:joy: :tada:

PR #187, opened 2020-02-03
deps: bump to is-wsl@2.2.0 

paulirish:
edit: ignore everything after the hr

just bringing over two small fixes. one for linux and one for docker. https://github.com/sindresorhus/is-wsl/releases

------------

wsl2 happened and is-wsl added support: https://github.com/sindresorhus/is-wsl/pull/3

ref https://github.com/GoogleChrome/lighthouse/issues/10288

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@types/node": "*",
-    "is-wsl": "^2.1.1",
+    "is-wsl": "^2.2.0",
     "lighthouse-logger": "^1.0.0",
     "mkdirp": "0.5.1",
     "rimraf": "^2.6.1"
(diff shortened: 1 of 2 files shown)


patrickhulce:
We should already have WSL 2 support @paulirish 

https://github.com/sindresorhus/is-wsl/releases/tag/v2.1.0

2.1.1 was just a [linux bug fix](https://github.com/sindresorhus/is-wsl/commit/08de510513156cee69bd9bcbae36cafca4d21cd0) for returning `false`

I think something else might be going on here...

paulirish:
ah it was done back here: https://github.com/GoogleChrome/chrome-launcher/pull/152

okay yeah definitely something else going on.

PR #191, opened 2020-03-16
feat: add waitForInspectableTarget option (fix #145)

### LINKED ISSUE #145: Occasional "No inspectable targets" error when launching chrome with a very short connectionPollInterval
To reproduce, create and run a Node script such as the one below:

```
const chromeLauncher = require('chrome-launcher');
const CDP = require('chrome-remote-interface');

let opts = {
  connectionPollInterval: 2,
  chromeFlags: ['--headless'],
};

(async() => {
  for (let i = 0; i < 100; i++) {
    let chrome = await chromeLauncher.launch(opts);
    let client = await CDP({port: chrome.port});
    console.log(i, chrome.port);
    client.close();
    chrome.kill();
  }
})().catch(error => { 
  console.log(error); 
  process.exit(1) 
});
```

After an unpredictable number of loops, the script (usually) crashes with a stacktrace:

```
Error: No inspectable targets
    at defaultTarget (.../node_modules/chrome-remote-interface/lib/chrome.js:47:23)
    at Chrome._fetchDebuggerURL (.../node_modules/chrome-remote-interface/lib/chrome.js:188:32)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
```

It looks like `chrome-launcher` occasionally reports that the launched Chrome instance is ready and has a debugging port, even though it isn't actually ready to accept a debug session yet.

gamtiq:
I've added new option `waitForInspectableTarget` that allows to wait until target list is available. It should fix #145.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ npm install chrome-launcher
   // Default: 50
   maxConnectionRetries: number;
 
+  // (optional) Interval in ms, which defines whether and how long an inspectable target should be awaited.
+  // `0` means that list of inspectable targets will not be requested and awaited.
+  // Default: 0
+  waitForInspectableTarget: number;
+
   // (optional) A dict of environmental key value pairs to pass to the spawned chrome process.
   envVars: {[key: string]: string};
 };
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
     "is-wsl": "^2.1.0",
     "lighthouse-logger": "^1.0.0",
     "mkdirp": "0.5.1",
+    "phin": "^3.4.1",
+    "povtor": "^1.1.0",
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -77,6 +77,11 @@ npm install chrome-launcher
   // Default: 50
   maxConnectionRetries: number;
 
+  // (optional) Interval in ms, which defines whether and how long an inspectable target should be awaited.
+  // `0` means that list of inspectable targets will not be requested and awaited.
+  // Default: 0
+  waitForInspectableTarget: number;
```
connorjclark, line 83: Wondering if this ought to be the default behavior. about to do a major version anyhow, so we can do a breaking change. Will play with this more later.

PR #197, opened 2020-05-07
Sanitise environment variables used in RegExps

Alhadis:
The `PATH` and `CHROME_PATH` variables are currently being used verbatim for constructing a regular expression:

https://github.com/GoogleChrome/chrome-launcher/blob/9aa64d642d3af13e447f8f84cacb05e556ff036e/src/chrome-finder.ts#L52-L53

This can lead to confusing behaviour if a user's path contains RegExp metacharacters such as `$` or `^`. Worse still, it leaves room for a malicious actor to inject an expression guaranteed to send the process into performance hell:

~~~shell
export CHROME_PATH='(.*.*.*.*.*.*.*.*.*.*.*.*)*[^Bb]|$'
~~~

I also replaced stuff like `` `${process.env.PATH}` `` with simply `process.env.PATH`, because the former syntax will stringify an undefined value as `"undefined"`, which probably isn't what you want (or expect).

### DIFF
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -7,8 +7,9 @@
 
 const fs = require('fs');
 const path = require('path');
-const execSync = require('child_process').execSync;
-const execFileSync = require('child_process').execFileSync;
+const {homedir} = require('os');
+const {execSync, execFileSync} = require('child_process');
+const escapeRegExp = require('escape-string-regexp');
 const log = require('lighthouse-logger');
 
 import {getLocalAppDataPath, ChromePathNotSetError} from './utils';
@@ -46,11 +47,13 @@ export function darwin() {
         });
       });
 
+
   // Retains one per line to maintain readability.
   // clang-format off
+  const home = escapeRegExp(process.env.HOME || homedir());
   const priorities: Priorities = [
-    {regex: new RegExp(`^${process.env.HOME}/Applications/.*Chrome.app`), weight: 50},
-    {regex: new RegExp(`^${process.env.HOME}/Applications/.*Chrome Canary.app`), weight: 51},
+    {regex: new RegExp(`^${home}/Applications/.*Chrome\\.app`), weight: 50},
+    {regex: new RegExp(`^${home}/Applications/.*Chrome Canary\\.app`), weight: 51},
     {regex: /^\/Applications\/.*Chrome.app/, weight: 100},
     {regex: /^\/Applications\/.*Chrome Canary.app/, weight: 101},
     {regex: /^\/Volumes\/.*Chrome.app/, weight: -2},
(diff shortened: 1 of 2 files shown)


src/chrome-finder.ts
```diff
@@ -202,7 +206,7 @@ function sort(installations: string[], priorities: Priorities) {
       .map(pair => pair.path);
 }
 
-function canAccess(file: string): Boolean {
+function canAccess(file?: string): Boolean {
```
patrickhulce, line 209:
```suggestion
function canAccess(file: string | undefined): Boolean {
```

we always want them to pass a file argument, but it's OK if it's undefined :)

src/chrome-finder.ts
```diff
@@ -141,19 +144,20 @@ export function linux() {
...
+    priorities.unshift({regex: new RegExp(escapeRegExp(process.env.CHROME_PATH)), weight: 101});
   }
 
   return sort(uniq(installations.filter(Boolean)), priorities);
 }
 
 export function wsl() {
   // Manually populate the environment variables assuming it's the default config
-  process.env.LOCALAPPDATA = getLocalAppDataPath(`${process.env.PATH}`);
+  process.env.LOCALAPPDATA = getLocalAppDataPath(<string>process.env.PATH);
```
patrickhulce, line 160:
```suggestion
  process.env.LOCALAPPDATA = getLocalAppDataPath(`${process.env.PATH}`);
```

this one seems necessary because PATH might not be defined, no?

Alhadis: If it's really undefined, then the expression `` `${process.env.PATH}` `` would yield the literal string `"undefined"`. That is, it's the same as this:

~~~js
"undefined" === `${undefined}`;
~~~

Which would match a directory named `./undefined` if it happened to contain `/Applications/Google Chrome.app`, etc. Though why a user would have `./undefined/Applications/Google Chrome.app` in their CWD is beyond me, but it's certainly not the intended behaviour... 😉

patrickhulce: Right, I know how string coercion works 😉 

I have a problem with unsafely casting this because `getLocalAppDataPath` is expecting a string and a `pathArgument.match` would *throw* instead of just generating a path with an `undefined` in it, which while confusing would *at worst* generate a different priority order if users are storing important programs in directories called `undefined` compared to fatally throwing every time `PATH` is `undefined`.

In this particular case with our *current* implementation, this change has no benefits with respect to your `undefined/Applications/` concern because both will not match the `regex.exec` so `undefined` will be forced into the path regardless. All this change does is expose us to future `cannot read property X` of undefined problems.

So I'd request either
- Properly fix getLocalAppDataPath to accept `undefined` as a type and handle it accordingly
- Revert the change because those wrinkles weren't part of what you were trying to fix and you've been very kind to help out enough already :)

Alhadis: Oh, right, I'm with you now. Sorry, it's late and I need sleep. 😅 (Now might be a good time to point out my TypeScript knowledge is rudimentary at best).

I've rolled it back.

patrickhulce: no worries, thanks! :)

PR #202, opened 2020-05-16
Update typescript & types

G-Rath:
Latest TypeScript is more accurate, faster, and mathematically higher 🥳

I also updated `@types/rimraf`, and `@types/sinon`: while this is bumping majors, it's raising them to match the current version of the package in use, as `@types` packages match the major.minor version of the package their typing (i.e so `@types/sinon@9` should be used for `sinon@9`, while `@types/sinon@7` should be used for `sinon@7`)

This revealed that the `this.timeout =` assignment being done in a test was invalid (since arrow functions don't have access to the mocha context); it's actually not needed because the timeout is set higher than 3500 via the launching flag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "devDependencies": {
     "@types/mkdirp": "^0.5.2",
     "@types/mocha": "^7.0.2",
-    "@types/rimraf": "^2.0.3",
+    "@types/rimraf": "^3.0.0",
     "@types/sinon": "^7.5.2",
     "clang-format": "^1.0.50",
     "mocha": "^7.1.1",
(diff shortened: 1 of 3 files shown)


G-Rath:
@patrickhulce I've opened up two minor PRs for now, because I don't want to tornado through your project tearing things up and stepping on toes; if you'd like I'm happy to go further, namely by enabling `strict` mode.

I did also notice that the `type-check` command is currently only checking `js` files, and doesn't use the same settings as your `tsconfig` - I'm happy to tweak that as well if you'd like 🙂

paulirish:
published in 0.13.3

PR #203, opened 2020-06-04
flags: disable background tracing

paulirish:
see 🔒  crbug.com/1058632 for details

it appears this affects canary/dev much more than stable. 

In Lighthouse we normally don't see this error in CLI cases where we have a fresh profile, but rather in devtools. But regardless, it makes sense to avoid background tracing regardless.

fixes (partially) https://github.com/GoogleChrome/lighthouse/issues/1091

### DIFF
--- a/src/flags.ts
+++ b/src/flags.ts
@@ -33,4 +33,6 @@ export const DEFAULT_FLAGS: ReadonlyArray<string> = [
   '--disable-renderer-backgrounding',
   // Disable task throttling of timer tasks from background pages.
   '--disable-background-timer-throttling',
+  // Disable background tracing (aka slow reports & deep reports) to avoid 'Tracing already started'
+  '--force-fieldtrials=*BackgroundTracing/default/',
 ];


patrickhulce:
sweet! I can't see the bug even on my chromium account, but sounds cool :)

paulirish:
@brendankenny wanna take a peek?

paulirish:
> I don't suppose there's some chromium doc on this we could link to in the comment? I'd not heard of `BackgroundTracing` before that crbug.

there's not a lot. these are the 3 best i could find:

* https://github.com/azharivs/chrome-tracing/wiki/How-to-trace-chrome-for-a-long-time-in-several-navigation%3F
* https://github.com/brave/brave-browser/issues/2121
* https://chromium.googlesource.com/chromium/src/+/8464ef03ddc445eb0041bfbdd15b5102c7eb9280

nothing great in here: https://google.com/search?q=chromium++%22slow+reports%22+OR+%22deep+reports%22 


i'd be fine with keeping the resources here in this PR thread since there's nothing standalone with enough context. sg?

Issue #207, opened 2020-06-08
Finding installation path is sometimes slow on mac


LarsDenBakker:
On my macbook pro it takes around 5 seconds to find my installation of chrome. This slows down execution of tests a lot. I asked around to some friends, and for some of them it's also this slow while for others it's fast.

I did some debugging in the code, and I think the culprit is here https://github.com/GoogleChrome/chrome-launcher/blob/master/src/chrome-finder.ts#L36 where the -dump listing is simply a lot of text. Unfortunately not something that's we can avoid.

Looking at the list of priorities here: https://github.com/GoogleChrome/chrome-launcher/blob/master/src/chrome-finder.ts#L54 I think we can speed up the lookup for the common cases by first doing a quick direct FS check fo these file paths. If they don't exist, then you can fallback on the more extensive lookup.

I can open a PR for this if this is something you think is worth adding.

patrickhulce:
Thanks for filing @LarsDenBakker this just started happening for me locally as well! I think it might be problematic to ensure we return the same output of `getInstallations` without it but I agree something should be done for this.

christian-bromann:
@patrickhulce do we want to always return all available Chrome apps? Otherwise we could just check for the default paths (e.g. `'/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'`) and return that if found by `fs.existsSync`.

Happy to make the change if you agree with this solution.

patrickhulce:
I think that change would be great for the default chrome-launcher path of `.launch`

i.e. we leave `getInstallations` alone but create `getFirstInstallation` that does exactly what you suggest without needing to invoke the -dump

christian-bromann:
> i.e. we leave `getInstallations` alone but create `getFirstInstallation` that does exactly what you suggest without needing to invoke the -dump

Are you suggesting that `.launch` could take a `preferDefault` parameter that if true, checks default installation paths and returns them if found or invokes -dump if not.

patrickhulce:
Not exactly. I was thinking of just a function kinda like

```js
function getFirstInstallation() {
  // list of the possibilities in priority order
  const defaultLocations = [
    process.env.CHROME_PATH,
    process.env.OTHER_CHROME_THING,
    '/Applications/Google Chrome.app/...',
    // ...
  ]
  for (const defaultLocation of defaultLocations) {
    if (canAccess(defaultLocation)) return defaultLocation
  } 

  return getInstallations()[0]
}
```

[...]

PR #209, opened 2020-06-25
Check default paths first when running on Mac

christian-bromann:
Running the lsregister dump can take up to a couple seconds on MacOS. This patch ensures that we check for default paths first and only do the dump if the browser is not located in `/Applications/Google Chrome.app` or `/Applications/Google Chrome Canary.app`.

Note: I couldn't really test all execution paths for the new method as I am not able to stub the `fs` package entirely (as you could do it with jest for example).

fixes: #207

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -177,7 +177,7 @@ class Launcher {
     return DEFAULT_FLAGS.slice();
   }
 
-  static getInstallations(): string[] {
+  static getInstallations() {
     return chromeFinder[getPlatform() as SupportedPlatforms]();
   }
 
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -8,6 +8,7 @@
 import * as childProcess from 'child_process';
 import * as fs from 'fs';
 import * as net from 'net';
+import * as path from 'path';
 import * as rimraf from 'rimraf';
 import * as chromeFinder from './chrome-finder';
 import {getRandomPort} from './random-port';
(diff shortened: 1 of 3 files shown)


src/chrome-launcher.ts
```diff
@@ -181,6 +182,24 @@ class Launcher {
     return chromeFinder[getPlatform() as SupportedPlatforms]();
   }
 
+  static getFirstInstallation(): string[] {
```
patrickhulce, line 185: I think this actually makes sense to live right next to getInstallations so we can keep the priorities in sync

maybe also limit the short-circuit to `os.platform() === 'darwin'`?

src/chrome-launcher.ts
```diff
@@ -181,6 +182,24 @@ class Launcher {
...
+    // list of the possibilities in priority order
+    const defaultLocations: string[] = [
+      process.env.CHROME_PATH as string, process.env.LIGHTHOUSE_CHROMIUM_PATH as string,
+      // Darwin
+      `${path.sep}Applications${path.sep}Google Chrome.app${path.sep}Contents${path.sep}MacOS${
```
patrickhulce, line 190: probably don't need `path.sep` since we'll do this darwin only

src/chrome-launcher.ts
```diff
@@ -181,6 +182,24 @@ class Launcher {
...
+          path.sep}Google Chrome`,
+      `${path.sep}Applications${path.sep}Google Chrome Canary.app${path.sep}Contents${
+          path.sep}MacOS${path.sep}Google Chrome Canary`
+    ].filter(Boolean)
+
+    for (const defaultLocation of defaultLocations) {
+      if (chromeFinder.canAccess(defaultLocation)) return [defaultLocation]
```
patrickhulce, line 197:
```suggestion
      if (!defaultLocation) continue;
      if (chromeFinder.canAccess(defaultLocation)) return defaultLocation;
```

src/chrome-launcher.ts
```diff
@@ -219,7 +238,7 @@ class Launcher {
       }
     }
     if (this.chromePath === undefined) {
-      const installations = Launcher.getInstallations();
+      const installations = Launcher.getFirstInstallation();
```
patrickhulce, line 241: feels like this API should just return the first, not an array

christian-bromann:
@patrickhulce addressed comments

src/chrome-launcher.ts
```diff
@@ -177,10 +177,33 @@ class Launcher {
     return DEFAULT_FLAGS.slice();
   }
 
-  static getInstallations() {
+  static getInstallations(): string[] {
+    // fetching the lsregister for all applications can be time consuming
+    // on MacOS therefor look into default paths first
+    if (getPlatform() === 'darwin') {
```
patrickhulce, line 183: Ah, I'm sorry for the miscommunication. I meant `getFirstInstallation` should live near the chrome finder logic that lists all of these darwin priorities. We should definitely leave `getInstallations` alone for this PR.

Sorry for my poor choice of words!

down below in launch we just want to use `getFirstInstallation()` *instead* of `getInstallations()[0]`

christian-bromann:
@patrickhulce thanks for the feedback. I moved the logic into the Chrome finder. Looks much better now.

PR #212, opened 2020-08-07
Fix getInstallations() section

weslenng:
`getInstallations()` is a method  of `Launcher` class


weslenng:
> @googlebot I signed it!

PR #218, opened 2020-10-31
Search in Path for chrome.exe

### LINKED ISSUE #214: Error "Runtime error encountered: No Chrome installations found."
#### Summary

On Windows 10, I use the Scoop package manager to install Chromium, Google Chrome, and Node.js. When I run `chrome.exe` from a PowerShell terminal, it launches Chromium 84, which includes lighthouse in the DevTools.

When I run `lighthouse` CLI, I received an error stating `Runtime error encountered: No Chrome installations found.`

It seems that something in how `lighthouse` CLI searches for Chrome installations is borked?

```
npm install --global lighthouse
lighthouse trevorsullivan.net
```
  
```
[cbt] C:\Users\TrevorSullivan\git> lighthouse.cmd trevorsullivan.net
Runtime error encountered: No Chrome installations found.
Error
    at new LauncherError (C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\utils.js:37:22)
    at new ChromeNotInstalledError (C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\utils.js:68:9)
    at Launcher.<anonymous> (C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\chrome-launcher.js:170:27)
    at Generator.next (<anonymous>)
    at C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\chrome-launcher.js:13:71
    at new Promise (<anonymous>)
    at __awaiter (C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\chrome-launcher.js:9:12)
    at Launcher.launch (C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\chrome-launcher.js:156:16)
    at Object.<anonymous> (C:\Users\TrevorSullivan\scoop\persist\nodejs\bin\node_modules\lighthouse\node_modules\chrome-launcher\dist\chrome-launcher.js:48:24)
    at Generator.next (<anonymous>)
```

lighthouse runs successfully

#### Environment Information
* Affected Channels: CLI
* Lighthouse version: 6.2.0
* Chrome version: chrome.exe points to Chromium 84.0.4147.125 (Official Build) (64-bit)
* Node.js version: v14.8.0
* Operating System: Windows 10 version 2004

**Related issues**

GoogleChrome/lighthouse#847

magneticflux-:
Allows finding Scoop installations, among others.

Closes #214

### DIFF
--- a/src/chrome-finder.ts
+++ b/src/chrome-finder.ts
@@ -197,12 +197,22 @@ export function win32() {
     installations.push(customChromePath);
   }
 
+  // Search common locations for chrome.exe
   prefixes.forEach(prefix => suffixes.forEach(suffix => {
     const chromePath = path.join(prefix, suffix);
     if (canAccess(chromePath)) {
       installations.push(chromePath);
     }
   }));
+
+  // Search %PATH% for chrome.exe
+  process.env.Path?.split(';')?.forEach(prefix => {
+    const chromePath = path.join(prefix, 'chrome.exe')
+    if (canAccess(chromePath)) {
+      installations.push(chromePath)
+    }
+  });
+
   return installations;
 }
 


magneticflux-:
@patrickhulce A `hacktoberfest-accepted` label would be much appreciated if this seems like a good PR! 😉

src/chrome-finder.ts
```diff
@@ -197,12 +197,22 @@ export function win32() {
...
+  // Search common locations for chrome.exe
   prefixes.forEach(prefix => suffixes.forEach(suffix => {
     const chromePath = path.join(prefix, suffix);
     if (canAccess(chromePath)) {
       installations.push(chromePath);
     }
   }));
+
+  // Search %PATH% for chrome.exe
+  process.env.Path?.split(';')?.forEach(prefix => {
```
connorjclark, line 209: - shouldn't this be `PATH`? Did it work for you as `Path`? Or are env vars in windows cased like that?
- Seems we could search PATH in mac (darwin in our code) and linux too
- I think the prefixes array about could have this added (`...process.env.Path.split(';')`), and this extra loop can be removed

connorjclark (changes requested):

src/chrome-finder.ts
magneticflux-: - I got confused between `Path` and `PATH` because of https://github.com/nodejs/node/issues/20605.
- I'll take a look at adding that too.
- Replaced the second loop.

magneticflux-: On Linux, `which` is used, so manual `PATH` parsing is unneeded.

On Darwin, `lsregister` is used, which _I think_ means it looks up apps:
> The Launch Service database controls the Finder's 'Open With' right click, contextual menu.

connorjclark: oh, found a reason to use `where` over doing it manually: in windows there's a way to use semicolons for individual items in the PATH... https://superuser.com/questions/584870/how-can-i-add-a-folder-containing-a-semicolon-to-the-windows-path . so if we use `where` we don't have to worry about the particulars of parsing PATH at all, which apparently isn't as trivial as unix (in unix/POSIX it is _impossible_ to escape the semicolon)

absir1949:
associated #277

connorjclark:
`which -a` changes still look good!

@adamraine will verify the window bits, then we can merge. Sorry for the delay @magneticflux-

connorjclark (approved):

adamraine:
I couldn't verify this, the command `where chrome.exe` did not return any results when I ran it on it's own on my machine

connorjclark:
> I couldn't verify this, the command `where chrome.exe` did not return any results when I ran it on it's own on my machine

It's because having chrome in PATH is not a typical setup, but it is a somewhat reasonable thing for an advanced user to configure. Can you add the folder where chrome is installed to the PATH, and then confirm that the code in chrome-finder.js is getting the correct path?

adamraine:
Ah thanks. Looks like this works as expected 👍

PR #221, opened 2020-12-01
Update dependencies and test targets

westy92:
* Update TypeScript and other dependencies.
* Update test targets. Current `LTS` and `stable` are `14.x` and `15.x` (https://nodejs.org/en/about/releases/).
* Cleanup code; update imports to improve type hints and remove deprecations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: node_js
 matrix:
   include:
-    - node_js: "8"
-    - node_js: "10"
-    - node_js: "12"
+    - node_js: 10
+    - node_js: 12
+    - node_js: lts/* # latest LTS release
+    - node_js: node # latest stable release
 sudo: required
 cache:
   yarn: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 matrix:
   include:
+    # Currently supported Node.js releases: https://nodejs.org/en/about/releases/
     - node_js: 10
     - node_js: 12
     - node_js: lts/* # latest LTS release
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 matrix:
   include:
     # Currently supported Node.js releases: https://nodejs.org/en/about/releases/
+    - node_js: 8
     - node_js: 10
     - node_js: 12
     - node_js: lts/* # latest LTS release
--- a/package.json
+++ b/package.json
@@ -11,22 +11,22 @@
     "prepublishOnly": "npm run build && npm run test"
   },
   "devDependencies": {
-    "@types/mkdirp": "^0.5.2",
-    "@types/mocha": "^7.0.2",
+    "@types/mkdirp": "^1.0.1",
+    "@types/mocha": "^8.0.4",
     "@types/rimraf": "^3.0.0",
     "@types/sinon": "^9.0.1",
     "clang-format": "^1.0.50",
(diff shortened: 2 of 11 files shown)


Niek (approved):

Niek:
@patrickhulce any chance to merge this?

.travis.yml
```diff
@@ -1,9 +1,11 @@
 language: node_js
 matrix:
   include:
-    - node_js: "8"
```
patrickhulce, line 4: let's leave node 8 in here until we officially drop support for it in our next major?

I'm with ya on dropping it, just don't want to publish a new major at this exact moment :)

package.json
```diff
@@ -11,22 +11,22 @@
...
+    "mocha": "^8.2.1",
     "sinon": "^9.0.1",
-    "ts-node": "8.3.0",
-    "typescript": "3.9.2"
+    "ts-node": "^9.0.0",
+    "typescript": "^4.1.2"
   },
   "dependencies": {
     "@types/node": "*",
-    "escape-string-regexp": "^1.0.5",
```
patrickhulce, line 26: can't use 3+ without node 10 only

package.json
```diff
@@ -11,22 +11,22 @@
...
+    "escape-string-regexp": "^4.0.0",
     "is-wsl": "^2.2.0",
     "lighthouse-logger": "^1.0.0",
-    "mkdirp": "^0.5.3",
```
patrickhulce, line 29: same, though it technically works at the moment.

patrickhulce (changes requested):
thanks @westy92 all of your stated goals are great 👍 

just don't think we want to introduce breaking changes parts of this right now. maybe file an issue for the collection of breaking changes we'll want to batch for the next major? :)

westy92:
I reverted the changes that require Node.js 10+. I'm happy to open another PR once this lands that drops Node.js 8.x support that can sit around until the project is ready for it.

src/chrome-launcher.ts
```diff
@@ -304,7 +304,7 @@ class Launcher {
   waitUntilReady() {
     const launcher = this;
 
-    return new Promise((resolve, reject) => {
+    return new Promise<{}>((resolve, reject) => {
```
patrickhulce, line 307: these could probably be better expressed as `Promise<void>` but it's not a big deal :)

patrickhulce (approved):
thanks for working on this @westy92 !

src/chrome-launcher.ts
westy92: I was going to, but these existing spots had `{}` so I left it:
https://github.com/GoogleChrome/chrome-launcher/blob/master/test/chrome-launcher-test.ts#L26
https://github.com/GoogleChrome/chrome-launcher/blob/master/src/chrome-launcher.ts#L50

patrickhulce: heh, yeah those are mistakes IMO, but nbd

westy92:
You're welcome! I addressed all the comments and it looks like all the tests passed. 🎉 Is there anything else I need to do to get this merged in?

patrickhulce:
nope I was just waiting on the tests and then forgot about it :) thanks @westy92 !

westy92:
> I'm happy to open another PR once this lands that drops Node.js 8.x support that can sit around until the project is ready for it.

Just to close the loop on this, here's my promised PR: https://github.com/GoogleChrome/chrome-launcher/pull/222

PR #222, opened 2020-12-03
Require Node.js 10+

westy92:
Drop support for Node.js 8, which is no longer supported: https://nodejs.org/en/about/releases/

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,7 +5,7 @@ version: "{build}"
 environment:
   fast_finish: true
   matrix:
-    - nodejs_version: "8"
+    - nodejs_version: 10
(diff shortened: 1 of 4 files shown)


brendankenny:
Should we drop support for node 10 as well? Node 12 is already the "Maintenance LTS" and 10 goes EoL in April.

https://nodejs.org/en/about/releases/

Would let us get rid of `rimraf` as a dep at some point, which would be kind of nice.

westy92:
@brendankenny IMO Node 10 support shouldn't be dropped until April 30th when it reaches EoL, but that's not my decision to make.

paulirish (approved):
sounds good!

PR #227, opened 2020-12-22
Add additional chrome flags

paulirish:
extra descriptions of these at https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md

### feels good
- `disable-component-update`.. this is huge, as I don't believe `disable-background-networking` was toggling this off.
- `disable-client-side-phishing-detection`. probably not a huge deal, but why not.
- `disable-ipc-flooding-protection`. Been default in pptr since 2018. https://github.com/puppeteer/puppeteer/commit/2a88690582cf01c8a338a972b352c34738e6bfdf seems good
- `password-store=basic` and `use-mock-keychain`. default in chromedriver/pptr.  see links in the markdown.

### not sure....

- `enable-automation`. most of the [benefits](https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md#test--debugging-flags) are great, but the added infobar caused some problems for webplatformtests: https://github.com/web-platform-tests/wpt/pull/6348. thoughts?

closes #23 (so old. made redundant, finally)

### DIFF
--- a/docs/chrome-flags-for-tools.md
+++ b/docs/chrome-flags-for-tools.md
@@ -111,7 +111,7 @@ All use cases are different, so you'll have to choose which flags are most appro
 * `--disable-desktop-notifications`
 * `--disable-features=TranslateUI`: renamed `TranslateUI` to `Translate` in [Sept 2020](https://chromium-review.googlesource.com/c/chromium/src/+/2404484).
 * `--disable-infobars`: [Removed April 2014](https://codereview.chromium.org/240193003)
-* `--disable-save-password-bubble`
+* `--disable-save-password-bubble`: [Removed May 2016](https://codereview.chromium.org/1978563002)
 * `--disable-translate`: [Removed April 2017](https://codereview.chromium.org/2819813002/) Used to disable built-in Google Translate service.
 * `--ignore-autoplay-restrictions`: [Removed December 2017](https://chromium-review.googlesource.com/#/c/816855/) Can use `--autoplay-policy=no-user-gesture-required` instead.
 * `--safebrowsing-disable-auto-update`: [Removed Nov 2017](https://bugs.chromium.org/p/chromium/issues/detail?id=74848#c26)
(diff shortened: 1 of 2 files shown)


brendankenny (approved):
> * `enable-automation`. most of the [benefits](https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md#test--debugging-flags) are great, but the added infobar caused some problems for webplatformtests: [web-platform-tests/wpt#6348](https://github.com/web-platform-tests/wpt/pull/6348). thoughts?

Seems like it could be ok, but do we have any tests sensitive enough to window size/positioning that we may get errors like they saw in https://groups.google.com/a/chromium.org/g/platform-predictability/c/k1I70ez7suM/m/1iskWQeKAgAJ ?

brendankenny:
> > * `enable-automation`. most of the [benefits](https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md#test--debugging-flags) are great, but the added infobar caused some problems for webplatformtests: [web-platform-tests/wpt#6348](https://github.com/web-platform-tests/wpt/pull/6348). thoughts?
> 
> Seems like it could be ok, but do we have any tests sensitive enough to window size/positioning that we may get errors like they saw in https://groups.google.com/a/chromium.org/g/platform-predictability/c/k1I70ez7suM/m/1iskWQeKAgAJ ?

running Lighthouse's `yarn smoke` with this branch of chrome-launcher had no errors, at least

brendankenny:
element rects and the full-page-screenshot also look properly aligned, so seems ok to me

paulirish:
> > * `enable-automation`. most of the [benefits](https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md#test--debugging-flags) are great, but the added infobar caused some problems for webplatformtests: [web-platform-tests/wpt#6348](https://github.com/web-platform-tests/wpt/pull/6348). thoughts?
> 
> Seems like it could be ok, but do we have any tests sensitive enough to window size/positioning that we may get errors like they saw in [groups.google.com/a/chromium.org/g/platform-predictability/c/k1I70ez7suM/m/1iskWQeKAgAJ](https://groups.google.com/a/chromium.org/g/platform-predictability/c/k1I70ez7suM/m/1iskWQeKAgAJ) ?

on second thought i don't think we want this.  i'd take it if it didnt add its own infobar.

but i just had a bunch that the infobar mucks with the screenshots and.. indeed it does.

image

that issue: https://github.com/GoogleChrome/lighthouse/issues/4451

PR #234, opened 2021-03-29
deps: replace mkdirp.sync with fs.mkdirSync

takenspc:
Node.js 10 LTS has recursive `fs.mkdirSync` so that we don't need mkdirp package.
https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_mkdirsync_path_options

I tested on Windows 10 64bit 20H2 since mkdirp is used only on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "prepublishOnly": "npm run build && npm run test"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


brendankenny (approved):
Great, thanks!

PR #237, opened 2021-04-14
move to minimum node 12; remove rimraf

brendankenny:
Node 10 will be EoL on April 30, just about the time we might possibly do another release of chrome-launcher, so seems ok to do now.

We should bump the next release to `0.14.0`, though.

With node 12 comes `fs.rmdir(path, {recursive: true})`, so we don't need rimraf anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-# Chrome Launcher [![Linux Build Status](https://img.shields.io/travis/GoogleChrome/chrome-launcher/master.svg)](https://travis-ci.org/GoogleChrome/chrome-launcher) [![Windows Build Status](https://img.shields.io/appveyor/ci/paulirish/chrome-launcher/master.svg)](https://ci.appveyor.com/project/paulirish/chrome-launcher/branch/master) [![NPM chrome-launcher package](https://img.shields.io/npm/v/chrome-launcher.svg)](https://npmjs.org/package/chrome-launcher)
-
+# Chrome Launcher [![GitHub Actions Status Badge](https://github.com/GoogleChrome/chrome-launcher/workflows/🛠/badge.svg)](https://github.com/GoogleChrome/chrome-launcher/actions) [![NPM chrome-launcher package](https://img.shields.io/npm/v/chrome-launcher.svg)](https://npmjs.org/package/chrome-launcher)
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,5 +1,4 @@
-# Chrome Launcher [![Linux Build Status](https://img.shields.io/travis/GoogleChrome/chrome-launcher/master.svg)](https://travis-ci.org/GoogleChrome/chrome-launcher) [![Windows Build Status](https://img.shields.io/appveyor/ci/paulirish/chrome-launcher/master.svg)](https://ci.appveyor.com/project/paulirish/chrome-launcher/branch/master) [![NPM chrome-launcher package](https://img.shields.io/npm/v/chrome-launcher.svg)](https://npmjs.org/package/chrome-launcher)
-
+# Chrome Launcher [![GitHub Actions Status Badge](https://github.com/GoogleChrome/chrome-launcher/workflows/🛠/badge.svg)](https://github.com/GoogleChrome/chrome-launcher/actions) [![NPM chrome-launcher package](https://img.shields.io/npm/v/chrome-launcher.svg)](https://npmjs.org/package/chrome-launcher)
```
brendankenny, line 1: drive-by badge fix for the migration to github actions

tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "compilerOptions": {
         "module": "commonjs",
         "outDir": "dist",
-        "target": "es2016",
+        "target": "es2019",
```
brendankenny, line 5: as officially suggested [here](https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-12). Savings in `dist/` are pretty small (about 3KiB), but it's nice that async/await is no loner transpiled, yielding a pretty big readability improvement in `chrome-launcher.js` if you aren't using source maps.

paulirish (approved):

lukebro:
In node v14.14+ a deprecation warning is now always given:

```
[DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

https://nodejs.org/api/deprecations.html#DEP0147

PR #242, opened 2021-06-14
fix: forwards support for fs.rm in node 14.14+

lukebro:
In node v14.14+ a deprecation warning is now always given:

[DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
nodejs.org/api/deprecations.html#DEP0147

So I've checked for the existence of `fs.rm` in favor of `fs.rmdir` if available, which should support v16+ but also v12.

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -377,8 +377,11 @@ class Launcher {
         this.fs.closeSync(this.errFile);
         delete this.errFile;
       }
-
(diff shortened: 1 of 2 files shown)


lukebro:
@googlebot I asigned it!

brendankenny (approved):
Sorry for the delay, LGTM. Would love to have this in for the next Lighthouse release.

Issue #213, opened 2020-08-08
[Feature Request] Ability to set preferences


Niek:
Some features can't be changed with Chrome flags, but require a manual entry in the data dir `Preferences` file. For example: the ``--lang`` flag is (no longer?) working on macOS, but ``intl.accept_languages`` preference works fine.

It would be great if `launch()` would have an `extraPrefs` or similar option to pass in an JSON object of preferences to put in the data dir `Preferences` file.

christian-bromann:
Created a PR adding this feature request as we need it in [WebdriverIO](https://webdriver.io) too.

VincentClair:
Actually, I can't make it work with `--lang` on Mac (Chrome 93.0.4577.63 (Mac OS 10.15.7)), FR language by default.
```
    customLaunchers: {
        ChromeEn: {
          base: 'Chrome',
          flags: ['--lang=en']
        },
    }
```

@Niek mentioned a manual entry in the data dir Preferences file. Where can i find it and how can I set it please ?  
Or have you another way to configure language (in order to wait for PR) ?

Niek:
@VincentClair you need to write the JSON Preferences file to the user data dir prior to starting Chrome. See https://niek.github.io/chrome-features/ (bottom of page) for a list of available settings.

PR #247, opened 2021-09-06
feat: Allow user to set browser preferences

christian-bromann:
implements: #213

In [WebdriverIO](https://webdriver.io/) we offer to run automated browser tests using WebDriver or CDP (via Puppeteer). When we start a Puppeteer browser we use this package as it gives us more options to set the browser session. However it is not possible to set a custom set of browser preferences, e.g. to define the default download directory. This patch adds this capability to the package. To create feature parity given that Chromedriver allows this, it would be great if we could add this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,11 @@ npm install chrome-launcher
   // Do note, many flags are set by default: https://github.com/GoogleChrome/chrome-launcher/blob/master/src/flags.ts
   chromeFlags: Array<string>;
 
+  // (optional) Additional preferences to be set in Chrome, for example: {'download.default_directory': __dirname}
+  // See: https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup
+  // Do note, if you set preferences when using your default profile it will overwrite these
+  prefs: {[key: string]: Object};
+
   // (optional) Close the Chrome process on `Ctrl-C`
   // Default: true
   handleSIGINT: boolean;
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ npm install chrome-launcher
   chromeFlags: Array<string>;
 
   // (optional) Additional preferences to be set in Chrome, for example: {'download.default_directory': __dirname}
-  // See: https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup
+  // See: https://chromium.googlesource.com/chromium/src/+/main/chrome/common/pref_names.cc
   // Do note, if you set preferences when using your default profile it will overwrite these
   prefs: {[key: string]: Object};
 
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -26,9 +26,12 @@ type SupportedPlatforms = 'darwin'|'linux'|'win32'|'wsl';
 
 const instances = new Set<Launcher>();
 
+type JSONLike =|{[property: string]: JSONLike}|readonly JSONLike[]|string|number|boolean|null;
+
 export interface Options {
   startingUrl?: string;
   chromeFlags?: Array<string>;
+  prefs?: Record<string, JSONLike>;
   port?: number;
   handleSIGINT?: boolean;
   chromePath?: string;
@@ -105,6 +108,7 @@ class Launcher {
   private chromePath?: string;
   private ignoreDefaultFlags?: boolean;
   private chromeFlags: string[];
+  private prefs: Record<string, JSONLike>;
   private requestedPort?: number;
   private connectionPollInterval: number;
   private maxConnectionRetries: number;
@@ -127,6 +131,7 @@ class Launcher {
     // choose the first one (default)
     this.startingUrl = defaults(this.opts.startingUrl, 'about:blank');
     this.chromeFlags = defaults(this.opts.chromeFlags, []);
+    this.prefs = defaults(this.opts.prefs, {});
     this.requestedPort = defaults(this.opts.port, 0);
     this.chromePath = this.opts.chromePath;
     this.ignoreDefaultFlags = defaults(this.opts.ignoreDefaultFlags, false);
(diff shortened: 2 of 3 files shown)


src/chrome-launcher.ts
```diff
@@ -26,9 +26,12 @@ type SupportedPlatforms = 'darwin'|'linux'|'win32'|'wsl';
 
 const instances = new Set<Launcher>();
 
+type ValidPrefTypes = string|boolean|number
```
patrickhulce, line 29: seems like the type is much more complex here. was it intentional to limit to only a few preferences?

src/chrome-launcher.ts
```diff
@@ -197,6 +202,11 @@ class Launcher {
     this.outFile = this.fs.openSync(`${this.userDataDir}/chrome-out.log`, 'a');
     this.errFile = this.fs.openSync(`${this.userDataDir}/chrome-err.log`, 'a');
 
+    // set preferences
+    if (Object.keys(this.prefs).length) {
+      this.setBrowserPrefs()
```
patrickhulce, line 207: nit: if we validate here, let's pass the validated prefs in, otherwise move validation to setBrowserPrefs :)

src/chrome-launcher.ts
```diff
@@ -206,6 +216,23 @@ class Launcher {
     this.tmpDirandPidFileReady = true;
   }
 
+  private setBrowserPrefs() {
+    const preferenceFile = `${this.userDataDir}/Preferences`;
+    try {
+      if (this.fs.existsSync(preferenceFile)) {
+        // overwrite existing file
+        const file = this.fs.readFileSync(preferenceFile, 'utf-8');
+        const content = JSON.parse(file.toString());
```
patrickhulce, line 225: nit: shouldn't need `toString` if using `utf-8`

patrickhulce (commented):
thanks @christian-bromann !

src/chrome-launcher.ts
christian-bromann: No, I actually thought the `Preference` file would have a flat design. I update the type to be JSON like.

test/chrome-launcher-test.ts
```diff
@@ -64,6 +64,37 @@ describe('Launcher', () => {
     assert.strictEqual(fs.rmdir.callCount, 0);
   });
 
+  it('allows to overwrite browser prefs', async () => {
+    const existStub = stub().returns(true)
+    const readFileStub = stub().returns(Buffer.from(JSON.stringify({ some: 'prefs' })))
```
patrickhulce, line 69: will probably need to adjust these stubs to match readFile behavior (string, not buffer)

src/chrome-launcher.ts
```diff
@@ -206,6 +213,28 @@ class Launcher {
     this.tmpDirandPidFileReady = true;
   }
 
+  private setBrowserPrefs() {
+    // don't set prefs if not defined
+    if (Object.keys(this.prefs).length) {
+      return
```
patrickhulce, line 219:
```suggestion
    if (!Object.keys(this.prefs).length) {
      return;
```

patrickhulce (approved):
LGTM when tests pass, thanks! :)

README.md
```diff
@@ -40,6 +40,11 @@ npm install chrome-launcher
   // Do note, many flags are set by default: https://github.com/GoogleChrome/chrome-launcher/blob/master/src/flags.ts
   chromeFlags: Array<string>;
 
+  // (optional) Additional preferences to be set in Chrome, for example: {'download.default_directory': __dirname}
+  // See: https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup
```
paulirish, line 44: viewvc links are some oldddd version of the chromium repo. you can see a 2014 date up there.

here's the modern equivalent of that page: https://chromium.googlesource.com/chromium/src/+/main/chrome/common/pref_names.cc

via the "view in gittiles" _Link_ from here: https://source.chromium.org/chromium/chromium/src/+/main:chrome/common/pref_names.cc


i'll change that and merge this in.

paulirish:
```suggestion
  // See: https://chromium.googlesource.com/chromium/src/+/main/chrome/common/pref_names.cc
```

christian-bromann: Thanks Paul!

christian-bromann:
Is there anything else I can do?

NavyMint:
Hi guys. @patrickhulce @paulirish sorry for disturbing you, do you guys know when we can expect this feature to be released?

paulirish:
@christian-bromann this preference file thing.. works, right? 
We've never used this technique, but if it works for you, that's great to hear. :)

christian-bromann:
@paulirish I tried to verify and it seems to not work anymore. Launching Chrome as following:

```
"/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary" --remote-debugging-port=61586 --user-data-dir=/var/folders/3_/pzc_f56j15vbd9z3r0j050sh0000gn/T/lighthouse.XXXXXXX.cofPsDof --enable-automation --disable-popup-blocking --disable-extensions --disable-background-networking --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-sync --metrics-recording-only --disable-default-apps --mute-audio --no-first-run --no-default-browser-check --disable-hang-monitor --disable-prompt-on-repost --disable-client-side-phishing-detection --password-store=basic --use-mock-keychain --disable-component-extensions-with-background-pages --disable-breakpad --disable-dev-shm-usage --disable-ipc-flooding-protection --disable-renderer-backgrounding --force-fieldtrials=*BackgroundTracing/default/ --enable-features=NetworkService,NetworkServiceInProcess --disable-features=site-per-process,TranslateUI,BlinkGenPropertyTrees --window-position=0,0 --window-size=1200,900 --auto-open-devtools-for-tabs --window-size=1600,1200 about:blank
```

While having a pref file written at ` /var/folders/3_/pzc_f56j15vbd9z3r0j050sh0000gn/T/lighthouse.XXXXXXX.cofPsDof/Preferences` with the contents of

```json
{
    "devtools.preferences.currentDockState": "\"bottom\"",
    "devtools.preferences.splitLocation": "\"50%\"",
    "devtools.preferences.panel-selectedTab": "\"console\""
}
```

Does not to have any effect anymore.

Junyan:
Try setting it to this file: `/var/folders/3_/pzc_f56j15vbd9z3r0j050sh0000gn/T/lighthouse.XXXXXXX.cofPsDof/Default/Preferences`

image

paulirish:
@christian-bromann yeah i think you're missing the `/Default/` in the pref file path. I just tried this out and it seems to work.

christian-bromann:
Thanks @paulirish , I will look into raising a PR with a fix for this.

PR #255, opened 2021-11-08
add method for printing found chrome path

brendankenny:
Using Lighthouse user flows, I find myself alternating between being fine with the version of Chrome shipped with puppeteer and needing the Canary installed on my machine, depending on what I'm doing, but it's awkward to set the path manually, and it's not portable (e.g. for sample code)

`chrome-launcher` is already set up to find Canary, so I've been wanting a way to be able to easily run something like

```js
const browser = await puppeteer.launch({
  headless: false,
  executablePath: chromeLauncher.getChromePath(),
});
```

I don't see any non-ugly way of reaching into `chrome-launcher` to do this (someone correct me if I'm wrong), so this PR exposes a function to do this.

Also adds a bin option as I've also wanted it on the command line, e.g. `CHROME_PATH="$(npm exec print-chrome-path)"`. Please suggest better approaches for that too, if there are any :)

### DIFF
--- a/bin/print-chrome-path.js
+++ b/bin/print-chrome-path.js
@@ -1,0 +1,16 @@
+#!/usr/bin/env node
+
+/**
+ * @license Copyright 2021 Google Inc. All Rights Reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+ */
+'use strict';
+
+/** @fileoverview Prints the Chrome path that chrome-launcher will use. */
+
+const {getChromePath} = require('../dist/chrome-launcher.js');
+
+const chromePath = getChromePath();
+process.stdout.write(chromePath);
+process.exit(0);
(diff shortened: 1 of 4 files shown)


adamraine:
This is better, but importing `chrome-launcher` everywhere in samples just to get the path to Canary doesn't seem ideal. Maybe we could add a `getChromePath` function to the FR API?

package.json
```diff
@@ -13,6 +13,9 @@
     "type-check": "tsc --allowJs --checkJs --noEmit --target es2019 *.js",
     "prepublishOnly": "npm run build && npm run test"
   },
+  "bin": {
+    "print-chrome-path": "bin/print-chrome-path.js"
```
paulirish, line 17: add this file plz :)

bin/print-chrome-path.js
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
```
connorjclark, line 1: I assume this what chmod +x'd, although I don't know if there is a way to confirm that in GH ui.

related, benefits to using shebang vs just setting the bin script to `node bin/print-chrome-path.js` ?

brendankenny: > I assume this was `chmod +x`'d, although I don't know if there is a way to confirm that in GH ui

npm/yarn does the chmod on install

> related, benefits to using shebang vs just setting the bin script to `node bin/print-chrome-path.js` ?

unlike `script` entries, `bin` entries are just filenames that are symlinked into `node_modules/.bin`, so need the shebang

connorjclark: TIL, thx!

connorjclark (approved):

PR #259, opened 2021-12-09
fix: always close file handles to stdout/stderr logs

### LINKED ISSUE #260: Error "operation not permitted, unlink chrome-err.log"
#### Environment Information
System: Windows 10
Node.js version: v14.18.1
Chrome version: 96.0.4664.55

When I use custom userDataDir, and remove it after chrome exit.

```
const customUserDataDir = 'D:\\.browser-temp\\chrome-4f32efbed74f445c';

const chrome = await launch({
  userDataDir: customUserDataDir
});

chrome.process.on('close', () => {
  if (fs.existsSync(customUserDataDir)) {
    const rm = fs.rmSync || fs.rmdirSync;
  
    rm(customUserDataDir)
  }
});
```

```
(node:3204) UnhandledPromiseRejectionWarning: Error: EPERM: operation not permitted, unlink 'D:\\.browser-temp\\chrome-4f32efbed74f445c\\chrome-err.log'
```

```
The outFile and errFile should be handled by fs.closeSync when chrome exit.
```

Junyan:
It should closeSync outFile and errFile when userDataDir is custom.

fixes #260

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -405,11 +405,6 @@ class Launcher {
 
   destroyTmp() {
     return new Promise<void>(resolve => {
-      // Only clean up the tmp dir if we created it.
-      if (this.userDataDir === undefined || this.opts.userDataDir !== undefined) {
-        return resolve();
-      }
-
       if (this.outFile) {
         this.fs.closeSync(this.outFile);
         delete this.outFile;
@@ -418,6 +413,11 @@ class Launcher {
       if (this.errFile) {
         this.fs.closeSync(this.errFile);
         delete this.errFile;
+      }
+
+      // Only clean up the tmp dir if we created it.
+      if (this.userDataDir === undefined || this.opts.userDataDir !== undefined) {
+        return resolve();
       }
 
       // backwards support for node v12 + v14.14+


src/chrome-launcher.ts
```diff
@@ -405,11 +405,6 @@ class Launcher {
...
   destroyTmp() {
     return new Promise<void>(resolve => {
-      // Only clean up the tmp dir if we created it.
-      if (this.userDataDir === undefined || this.opts.userDataDir !== undefined) {
-        return resolve();
-      }
-
       if (this.outFile) {
         this.fs.closeSync(this.outFile);
         delete this.outFile;
```
connorjclark, line 410: Makes sense we should always close these file handles when done. I'm curious though, what was the error you were seeing?

connorjclark (approved):

connorjclark:
Can you run `yarn format`, or allow upstream to push to your fork branches so I can run it?

Junyan:
@connorjclark done

kiqaps:
guys, this still let error file open when used custom data dir, no?

i had to create a patch-package which moves the check condition (if we are using custom data dir) bellow the closeSync of errFile and it worked...

PR #265, opened 2022-05-29
add lighthouse smoketests (windows, lantern) to CI

paulirish:
per https://github.com/GoogleChrome/lighthouse/issues/14023 and https://github.com/GoogleChrome/lighthouse/issues/13871 we've been having some challenges with the windows smoke tests

and generally the problem is the `taskkill` command to close the browser.

I've been doing some work on this but first thing that makes sense is to add the LH smoketests (just windows) to the chrome-launcher CI.

typically its the lantern smoketests that trigger the failure, so i'm running just those.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "mocha": "^8.2.1",
     "sinon": "^9.0.1",
     "ts-node": "^9.1.0",
-    "typescript": "^4.1.2"
+    "typescript": "^4.1.2",
+    "lighthouse": "https://github.com/GoogleChrome/lighthouse.git#master"
   },
   "dependencies": {
     "@types/node": "*",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "test-formatting": "test/check-formatting.sh",
     "format": "scripts/format.sh",
     "type-check": "tsc --allowJs --checkJs --noEmit --target es2019 *.js",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -23,7 +24,8 @@
     "mocha": "^8.2.1",
     "sinon": "^9.0.1",
     "ts-node": "^9.1.0",
-    "typescript": "^4.1.2"
+    "typescript": "^4.1.2",
+    "lighthouse": "https://github.com/GoogleChrome/lighthouse.git#master"
```
connorjclark, line 28: FYI this doesn't actually get the _latest_ code, but it pinned in yarn.lock. If you want the latest, we can do yarn add -D in GH actions.

If using the latest doesn't particularly matter, then let's just pin to 9.6.1

paulirish: hmm yarn add -D in the yml would actually be preferable to me...  yah good call about the lock. 

i'll switch to this.

paulirish:
I'm happy with these. I'm reliably seeing the taskkill failures in the CI here. example run: https://github.com/GoogleChrome/chrome-launcher/runs/6662411663?check_suite_focus=true

The smokehouse retries often end up resolving it, so the CI often passes, but the above run was a case where `lantern-online` continued to fail despite 5 retries.

connorjclark (approved):

PR #267, opened 2022-05-30
revise taskkill procedure on windows

### LINKED ISSUE #266: Unreliable process killing on windows
There are several issues about this problem. (in this repo and in ligththouse's)

I'm going to jot down some notes here.

We use `taskkill` to kill the chrome process on windows:

https://github.com/GoogleChrome/chrome-launcher/blob/279577fd3fd4dda8a79b6657a43dbe42b2b3d7cb/src/chrome-launcher.ts#L388

We have intermittent failures where this command fails and thus `launcher.kill()` rejects.  The failures look like:

```
Error: Couldn't quit Chrome process. 
Error: Chrome could not be killed Command failed: taskkill /pid 13216 /T /F
ERROR: The process with PID 8536 (child process of PID 13216) could not be terminated.
Reason: The operation attempted is not supported.
```

History of related changes:

* https://github.com/GoogleChrome/chrome-launcher/pull/179 made sure stdio from the `taskkill` cmd is piped up but **not** emitted.
* https://github.com/GoogleChrome/chrome-launcher/pull/112 added the rejection, primarily for the sigINT usecase?

paulirish:
mostly fixes #266  

#### adopted much of playwright's logic ([detailed here](https://github.com/GoogleChrome/chrome-launcher/issues/266#issuecomment-1141492721))

* chrome process not detached on windows
* taskkill is done spawnsync instead of execsync. `shell:true` swapped in for `stdio:pipe`
* no chance of rejection in `kill()` matching pptr & pw behavior.

**also:**

* dropped our LH smoke retries to ZERO so the failures are loud and clear
* rename `this.chrome` to `this.chromeProc` (sorry for the noise, but we have another `chrome` in the public API that's a different thing)

(currently on top of #265 ... https://github.com/GoogleChrome/chrome-launcher/compare/lighthouse-smoke...adopt-pw-kill )


src/chrome-launcher.ts
```diff
@@ -373,28 +379,30 @@ class Launcher {
...
-            if (this.chrome.pid) {
-              process.kill(-this.chrome.pid);
+            if (this.chromeProc.pid) {
+              process.kill(-this.chromeProc.pid, 'SIGKILL');
             }
           }
         } catch (err) {
           const message = `Chrome could not be killed ${err.message}`;
           log.warn('ChromeLauncher', message);
-          reject(new Error(message));
```
adamraine, line 397: The smoke tests still pass even with this rejection right?

paulirish: no they don't. 

removing this rejection is deliberate actually. as mentioned in #266, pptr and pw deliberately do not reject in this case. 

and i've been convinced that this is reasonable. there appears to be a race condition and taskkill is attempting to kill processes that are already gone. what's interesting is that when this happens. the processes that fail to be killed (cuz they're not found) are all child proc of the primary chrome proc.  the chrome proc is always successfully killed.

take a peek at the SO item linked in https://github.com/microsoft/playwright/pull/7500 .... PW used this `MEMUSAGE` hack for a while but then determined it wasnt needed. so i'm following their latest pattern.

i think it's still important we have some visibility on when taskkill returns an error, but it's not a fatal situation.

adamraine (approved):

src/chrome-launcher.ts
```diff
@@ -276,9 +277,9 @@ class Launcher {
 
   private async spawnProcess(execPath: string) {
     const spawnPromise = (async () => {
-      if (this.chrome) {
-        log.log('ChromeLauncher', `Chrome already running with pid ${this.chrome.pid}.`);
-        return this.chrome.pid;
+      if (this.chromeProc) {
```
connorjclark, line 280: nit: chromeProcess

I'm reading "chromeProc" as "chromeProcedure" (because that is a weird/old way to shorten that word) so reading was confusing

PR #268, opened 2022-05-31
make launcher.kill() synchronous

paulirish:
per #266 there's no reason for this to be async.  keeping it async adds extra challenges to both debugging and understanding the flow.

also, added a function test for userdatadir removal

for review, turn off whitespace

### DIFF
--- a/src/chrome-launcher.ts
+++ b/src/chrome-launcher.ts
@@ -54,8 +54,8 @@ export interface ModuleOverrides {
   spawn?: typeof childProcess.spawn;
 }
 
-const sigintListener = async () => {
-  await killAll();
+const sigintListener = () => {
+  killAll();
   process.exit(_SIGINT_EXIT_CODE);
 };
 
@@ -92,11 +92,11 @@ function getChromePath(): string {
   return installation;
 }
 
-async function killAll(): Promise<Array<Error>> {
+function killAll(): Array<Error> {
   let errors = [];
   for (const instance of instances) {
     try {
-      await instance.kill();
+      instance.kill();
       // only delete if kill did not error
       // this means erroring instances remain in the Set
       instances.delete(instance);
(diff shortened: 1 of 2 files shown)


src/chrome-launcher.ts
```diff
@@ -380,60 +380,56 @@ class Launcher {
...
 
-      // backwards support for node v12 + v14.14+
-      // https://nodejs.org/api/deprecations.html#DEP0147
-      const rm = this.fs.rm || this.fs.rmdir;
-      rm(this.userDataDir, {recursive: true}, () => resolve());
-    });
+    // backwards support for node v12 + v14.14+
+    // https://nodejs.org/api/deprecations.html#DEP0147
+    const rmSync = this.fs.rmSync || this.fs.rmdirSync;
+    rmSync(this.userDataDir, {recursive: true, force: true});
```
connorjclark, line 432: why set force: true now?

connorjclark (approved):

src/chrome-launcher.ts
paulirish: if this dir doesn't exist, we want this fn call to not throw.

src/chrome-launcher.ts
```diff
@@ -380,60 +380,56 @@ class Launcher {
...
 
-      // backwards support for node v12 + v14.14+
-      // https://nodejs.org/api/deprecations.html#DEP0147
-      const rm = this.fs.rm || this.fs.rmdir;
-      rm(this.userDataDir, {recursive: true}, () => resolve());
-    });
+    // backwards support for node v12 + v14.14+
+    // https://nodejs.org/api/deprecations.html#DEP0147
+    const rmSync = this.fs.rmSync || this.fs.rmdirSync;
+    rmSync(this.userDataDir, {recursive: true, force: true, maxRetries: 10});
```
connorjclark, line 432: can you also capture the force / max retry change in the commit title

connorjclark: oops

calebeby:
@paulirish would it make sense to remove the [`async` modifier from the `kill` function returned by `launch()`](https://github.com/GoogleChrome/chrome-launcher/blob/master/src/chrome-launcher.ts#L75) too? Happy to send a PR!

paulirish:
@calebeby yup sg!

PR #269, opened 2022-06-09
Make LaunchedChrome.kill sync

calebeby:
@paulirish https://github.com/GoogleChrome/chrome-launcher/pull/268#issuecomment-1151335466


calebeby:
This is _technically_ a breaking change if someone was calling `.then()` on `.kill()` instead of using `await`

PR #275, opened 2022-07-27
remove option to start headless via env variable

markusmeyer:


paulirish:
why tho?

connorjclark:
@paulirish there is a referenced issue above your comment

paulirish:
@connorjclark ah thats true. didnt see that.

PR #286, opened 2022-12-02
use updating node versions in CI

paulirish:
..... 

okay it it looks happy now :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,16 +10,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # awaiting support for 'latest'/'lts'. https://github.com/actions/setup-node/issues/26
-        node: [ '12', '14', '16', '17' ]
+        node: [ 'latest', 'lts/*', 'lts/-n' ]
     name: basics (node ${{ matrix.node }})
 
     steps:
     - name: git clone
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -10,20 +10,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # awaiting support for 'latest'/'lts'. https://github.com/actions/setup-node/issues/26
-        node: [ '12', '14', '16', '17' ]
+        node: [ 'latest', 'lts/*', 'lts/-1' ]
```
brendankenny, line 13: One (small) advantage of the explicit version numbers is it matches what's in the package.json `engines` field. Usually they naturally get updated together (the 12 tests will fail if you update package.json but not the test matrix), but in this world these versions could keep advancing and `engines` could bit rot if nobody files an issue (and we don't release a breaking change but stop testing on older node versions).

One option is keeping an explicit oldest version in the matrix? Annoying if it duplicates a version, though. Maybe more important on the LH repo since it'll be testing chrome-launcher on its full version matrix.

brendankenny (approved):

.github/workflows/ci.yml
paulirish: true

Issue #299, opened 2023-05-16
/Users/sample/node_modules/chrome-launcher/dist/chrome-finder.js:14 const log = require('lighthouse-logger');      ^  Error [ERR_REQUIRE_ESM]: require() of ES Module


amrsa1:
Recently i start getting the following error when executing my test using webdriverIO

```
/Users/runner/work/sample/sample/node_modules/chrome-launcher/dist/chrome-finder.js:14
const log = require('lighthouse-logger');
            ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/runner/work/sample/sample/node_modules/lighthouse-logger/index.js from /Users/runner/work/sample/sample/node_modules/chrome-launcher/dist/chrome-finder.js not supported.
Instead change the require of index.js in /Users/runner/work/sample/sample/node_modules/chrome-launcher/dist/chrome-finder.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/runner/work/sample/sample/node_modules/chrome-launcher/dist/chrome-finder.js:14:13)
    at Object.<anonymous> (/Users/runner/work/sample/sample/node_modules/chrome-launcher/dist/chrome-launcher.js:11:22)
    at Object.<anonymous> (/Users/runner/work/sample/sample/node_modules/chrome-launcher/dist/index.js:13:14) {
  code: 'ERR_REQUIRE_ESM'
}
```

The error message points out that the require() statement in the file  chrome-launcher/dist/chrome-finder.js is trying to import an ES module (index.js from lighthouse-logger), but it is not supported in that context.

using dyanmic import() might solve the issue

adamraine:
See https://github.com/GoogleChrome/lighthouse/issues/15085

PR #300, opened 2023-05-16
#299: Fix ES Module require() not supported - Chrome Launcher and Lighthouse Logger conflict

amrsa1:
#299 ES Module require() not supported - Chrome Launcher and Lighthouse Logger conflict


adamraine:
Hi @amrsa1, cjs support has been restored in `lighthouse-logger` and the broken version has been deprecated

PR #302, opened 2023-05-16
Publish as ESM

adamraine:
This needs to be done in a breaking release, because this package is currently published as CJS and this PR changes the published JS to be ESM.

This will ensure Lighthouse and Chrome launcher are using the same version of lighthouse logger in their shared `lighthouse-logger` package.


connorjclark (approved):

test/chrome-launcher-test.ts
```diff
@@ -38,7 +41,7 @@ const launchChromeWithOpts = async (opts: Options = {}) => {
   }
 };
 
-describe('Launcher', () => {
+describe.only('Launcher', () => {
```
brendankenny, line 44: from debugging?

tsconfig.json
```diff
@@ -10,12 +11,14 @@
...
+        "noUnusedParameters": true,
+        "allowSyntheticDefaultImports": true,
     },
     "exclude": [
         "node_modules"
     ],
     "include": [
-        "src"
+        "src",
+        "types"
```
brendankenny, line 22: does it really need the d.ts file of itself?

test/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../tsconfig.json",
+  "ts-node": {
+    "files": true,
+  },
+  "include": [
+    "*-test.ts",
+    "../types"
+  ]
+}
```
brendankenny, line 10: nit: newline

test/chrome-launcher-test.ts
adamraine: Explains why the tests were passing

tsconfig.json
adamraine: Of itself? The d.ts file is for lighthouse logger

brendankenny: lol, I've been looking too long at lighthouse-logger I guess :)

PR #310, opened 2023-07-18
1.0.0

adamraine:
This needs to be a breaking rls because of https://github.com/GoogleChrome/chrome-launcher/pull/302


paulirish (approved):
big day

PR #314, opened 2023-08-30
docs(flags): document setting preferences

paulirish:


docs/chrome-flags-for-tools.md
```diff
@@ -180,6 +180,24 @@ Here's a **[Nov 2022 comparison of what flags](https://docs.google.com/spreadshe
...
+If you wanted to launch a fresh Chrome profile **_with_** some Preferences set, for example: disable DevTools source-map fetching:
+
+```sh
+mkdir -p your_empty_user_data_dir/Default/
+touch    your_empty_user_data_dir/Default/Preferences
+```
+
+Set the file content to:
+```json
+{"devtools":{"synced_preferences_sync_disabled":{"jsSourceMapsEnabled":"false","cssSourceMapsEnabled":"false"}}}
```
brendankenny, line 196: is the best reference for these just browsing your own profile or is there a good place listing them?

brendankenny (approved):

docs/chrome-flags-for-tools.md
paulirish: good call. added link to https://source.chromium.org/chromium/chromium/src/+/main:chrome/common/pref_names.h

PR #316, opened 2023-09-01
docs(flags): --disable-infobars fix link to removal CL

mathiasbynens:
The previously linked CL seems unrelated.


paulirish:
ah true. yeah the other CL did 'move' the flag..  my mistake. thanks!

paulirish (approved):

PR #317, opened 2023-09-01
Fix chrome-out.log and chrome-err.log keep opening after browser closed

datnv9:


paulirish (approved):

src/chrome-launcher.ts
```diff
@@ -356,6 +356,12 @@ class Launcher {
         };
       }
 
+      // handle process closed right after it started
+      this.chromeProcess.on('close', () => {
```
paulirish, line 360: this change fails some of the tests. not sure why at the moment.

But it may need to move slightly.

paulirish (changes requested):
the changes in destroyTmp() make complete sense. but i'm unsure about the event listener.   do you know whats going on there?

PR #319, opened 2023-09-29
add portStrictMode option

connorjclark:
I need this to ensure that I am connecting to a mobile device, and to fail if that expected port is not available (as opposed to silently fallback to launching Chrome on the current device).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ npm install chrome-launcher
   // Default: an available port is autoselected
   port: number;
 
-  // (optional) When a port is specified but no Chrome is found at that port:
-  // * if this option is true and there is no Chrome with that debug port, throw an error
-  // * else, launch a new Chrome with that debug port
+  // (optional) When `port` is specified *and* no Chrome is found at that port,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,13 @@ npm install chrome-launcher
   // Default: an available port is autoselected
   port: number;
 
+  // (optional) When a port is specified but no Chrome is found at that port:
+  // * if this option is true and there is no Chrome with that debug port, throw an error
+  // * else, launch a new Chrome with that debug port
```
paulirish, line 40:
```suggestion
  // (optional) When `port` is specified *and* no Chrome is found at that port,
  // * if `false` (default), chrome-launcher will launch a new Chrome with that port.
  // * if `true`, throw an error
```

paulirish (commented):
naming wise i don't love this one.. but.. as mentioned on chat... nothing is immediately great.

`portStrictMode` is the best ive got... lets do that ? sure.

paulirish (approved):

PR #331, opened 2024-05-22
flags: disable in-product health system

adamraine:
This disables the annoying UI help bubbles that appear when Chrome is opened "for the first time" which actually means every time for our setup.

See b/340895256 for more context.


paulirish:
This is amazing. Great find.

Issue #355, opened 2025-09-24
Windows Lighthouse error: Error: EPERM, Permission denied: ... temp directory ... at rmSync ... at Launcher.destroyTmp


AndreiTelteu:
Hi. I tried using lighthouse-mcp on windows and it crashes when getting the result. The  results are generated in a temporary folder but it crashes after "LH:status Generating results..."

Here is the last lines of the log with the stacktrace of the error:
```log
2025-09-24 22:35:38.347 | [info] MCP [pid: 95860] stderr: Lighthouse MCP server running on stdio
2025-09-24T19:34:28.971Z LH:config:warn unrecognized category in 'onlyCategories': pwa
2025-09-24T19:34:29.048Z LH:status Connecting to browser
2025-09-24T19:34:29.053Z LH:status Navigating to about:blank
2025-09-24T19:34:29.057Z LH:status Benchmarking machine
2025-09-24T19:34:30.061Z LH:status Host device pixel ratio
2025-09-24T19:34:30.061Z LH:status Preparing target for navigation mode
2025-09-24T19:34:30.073Z LH:status Cleaning origin data
2025-09-24T19:34:30.084Z LH:status Cleaning browser cache
2025-09-24T19:34:30.118Z LH:status Preparing network conditions
2025-09-24T19:34:30.156Z LH:status Navigating to https:// - redacted - 
2025-09-24T19:34:33.781Z LH:status Getting artifact: DevtoolsLog
2025-09-24T19:34:33.781Z LH:status Getting artifact: Trace
2025-09-24T19:34:33.781Z LH:status Getting artifact: Accessibility
2025-09-24T19:34:33.906Z LH:status Getting artifact: AnchorElements
2025-09-24T19:34:33.970Z LH:method <= browser ERR:error DOM.pushNodeByPathToFrontend 
2025-09-24T19:34:33.971Z LH:method <= browser ERR:error DOM.pushNodeByPathToFrontend 
..... a lot ......
2025-09-24T19:35:38.017Z LH:status Auditing: Optimize DOM size
2025-09-24T19:35:38.018Z LH:status Auditing: Duplicated JavaScript
2025-09-24T19:35:38.018Z LH:status Auditing: Font display
2025-09-24T19:35:38.019Z LH:status Auditing: Forced reflow
2025-09-24T19:35:38.019Z LH:status Auditing: Improve image delivery
2025-09-24T19:35:38.020Z LH:status Auditing: INP breakdown
[...]

PR #356, opened 2025-10-25
fix: Removed call to destroyTmp before chrome is killed.

hubuk:
Fix for #355
Current code calls ```destroyTmp``` before chrome process is killed for good.
If for some reasons that I am not aware of this call is needed it would be a good idea to put it inside try/catch with log instead.


hubuk:
CLA signed now.

